PR #2, opened 2026-03-12
Remove outdated #![allow(dead_code)] attributes

kitaebot:
## Summary

Removes module-level `#![allow(dead_code)]` attributes from three files that no longer need them.

## Changes

- `src/context.rs` - Removed `#![allow(dead_code)]`
- `src/error.rs` - Removed `#![allow(dead_code)]`  
- `src/types.rs` - Removed `#![allow(dead_code)]`

## Rationale

These attributes were originally added with comments indicating the code would be "used in later commits". All three modules are now fully utilized throughout the codebase:

- **context.rs**: Context window management with compaction logic, used by the agent loop
- **error.rs**: Core error types (`Error`, `ProviderError`, `ToolError`, etc.) used everywhere
- **types.rs**: Domain types (`Message`, `ToolCall`, `Response`, `ToolDefinition`, etc.) - the core protocol types

Removing these attributes ensures the compiler will properly warn about genuinely unused code in the future, rather than suppressing warnings for code that is actually in active use.

## Testing

These are purely cleanup changes - no functional changes. The code compiles and all existing tests should continue to pass.

### DIFF
--- a/src/context.rs
+++ b/src/context.rs
@@ -1,5 +1,3 @@
-#![allow(dead_code)]
-
 //! Context window management.
 //!
 //! Estimates token usage and compacts the session when it exceeds the
--- a/src/error.rs
+++ b/src/error.rs
@@ -3,8 +3,6 @@
 //! Models all failure modes explicitly as algebraic data types.
 //! No generic "something failed" errors - each error explains what went wrong.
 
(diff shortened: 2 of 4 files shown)


amackillop (changes requested):
Squash these commits into one and sign it properly

amackillop (approved):

PR #5, opened 2026-07-29
Heal wedged checkouts in git_clone too

amackillop:
ensure_cloned recently learned to detect and replace a corrupt .git skeleton (the wedged-clone incident from the warm duty's first live run), but the git_clone tool never benefited: its existing-checkout branch trusts is_dir(".git") and goes straight to fetch. A wedged checkout in the interactive path fails that fetch loudly but never heals.

This routes git_clone through ensure_cloned unconditionally. ensure_cloned returns an Ensured enum saying whether a clone ran or a healthy checkout was reused, so the tool keeps its two honest messages and fetches only in the reuse case, where the working tree may hold in-progress work and must never be repositioned or cleaned. A wedged skeleton is replaced and reported as a fresh clone, because that is what it is.

### DIFF
--- a/src/channel/github/review_checkout.rs
+++ b/src/channel/github/review_checkout.rs
@@ -67,7 +67,9 @@ async fn prepare_at(
     // Reviewing a repo the bot has never worked on clones it. Fetch
     // into the clone, not the worktree: they share refs, and the
     // worktree may not exist yet.
-    let clone = checkout::ensure_cloned(git, url, clone_rel).await?;
+    let clone = checkout::ensure_cloned(git, url, clone_rel)
+        .await?
+        .into_dir();
     let pull_ref = format!("pull/{pr_number}/head");
     checkout::run(git, &["fetch", "origin", base, &pull_ref], &clone, true).await?;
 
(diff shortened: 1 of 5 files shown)


kitaebot (approved):
Reviewed the diff against the commit message claims. All verified:

- `git_clone` is now routed through `ensure_cloned` unconditionally; the old `is_dir(".git")` short-circuit is gone.
- `ensure_cloned` returns an `Ensured` enum (`Cloned` / `Existing`) so the tool keeps its two honest messages and fetches only in the reuse case — a fresh clone is not re-fetched.
- The reuse path only fetches; it never repositions or cleans the working tree, preserving in-progress work. The invariant is stated in a comment.
- A wedged `.git` skeleton (directory present but `rev-parse --git-dir` fails) is removed and re-cloned, reported as `Cloned` — which is what it is.
- The three other callers (`review_checkout.rs`, `execution_checkout.rs`, `git_cli.rs`) collapse to a one-word `.into_dir()` change each, which is all they ever wanted.

Enum variant and match-arm order are alphabetical, comments are terse, and the `Ensured` enum makes the clone-vs-reuse distinction unrepresentable incorrectly. No findings.

PR #6, opened 2026-07-30
Trim direnv cache lookups to one stat

amackillop:
Small perf trim on the exec hot path: the direnv cache fingerprint costs two stat calls per lookup; one suffices. Rationale in the commit message.

### DIFF
--- a/src/tools/direnv.rs
+++ b/src/tools/direnv.rs
@@ -5,9 +5,9 @@
 //!
 //! # Cache invalidation
 //!
-//! Two `stat()` calls per lookup: `.envrc` mtime and `flake.lock` mtime.
-//! A changed mtime triggers re-evaluation. Failures are never cached — the
-//! next caller retries.
+//! One `stat()` call per lookup: the `.envrc` mtime. A changed mtime
+//! triggers re-evaluation. Failures are never cached — the next caller
+//! retries.
 //!
 //! # Concurrency
 //!
@@ -69,14 +69,12 @@ pub async fn allow(dir: &Path) {
 #[derive(Clone, Debug, PartialEq, Eq)]
 struct Fingerprint {
     envrc_mtime: Option<SystemTime>,
-    flake_lock_mtime: Option<SystemTime>,
 }
 
 impl Fingerprint {
     fn of(dir: &Path) -> Self {
         Self {
             envrc_mtime: mtime(&dir.join(".envrc")),
-            flake_lock_mtime: mtime(&dir.join("flake.lock")),
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/tools/direnv.rs
```diff
@@ -69,14 +69,12 @@ pub async fn allow(dir: &Path) {
 #[derive(Clone, Debug, PartialEq, Eq)]
 struct Fingerprint {
     envrc_mtime: Option<SystemTime>,
```
kitaebot, line 71: Dropping `flake.lock` from `Fingerprint` means a lock-only change (e.g. `nix flake update`) no longer invalidates the cache: the agent keeps running in the stale devshell until someone touches `.envrc`. The commit's rationale covers trust (direnv's allow hash is on `.envrc` content) but not this staleness case.

If the trade-off is intended, one line in the commit message acknowledging the known stale-window would close the gap. Alternatively, keying on `max(envrc_mtime, flake.lock_mtime)` only when `flake.lock` exists would preserve lock-change invalidation at the cost of a conditional second stat — still one stat on the common path where `flake.lock` is absent.

kitaebot (commented):
The code change is minimal and matches the commit message: `Fingerprint` drops `flake_lock_mtime` and its stat call, the module doc comment is updated in lockstep, and the now-meaningless `fingerprint_with_both_files` test is removed. No dead code left behind.

Two should-fix findings before this lands:

**1. Stale spec — `specs/03-tools.md` (not in the diff, so called out here).** The spec still contradicts the code in two places:
- Requirement 2 (line 431): "modified `.envrc` or `flake.lock` triggers re-evaluation"
- Invalidation section (lines 446–447): "Staleness is determined by the mtime of `.envrc` and `flake.lock` — two `stat` calls per lookup."

AGENTS.md says "Specs are contracts. When code diverges from a spec, fix the spec." This PR updates the module doc comment but leaves the governing spec describing the old two-file invalidation. Both passages should be updated to `.envrc`-only in this commit.

**2. flake.lock-only staleness window (inline comment on direnv.rs).**

src/tools/direnv.rs
amackillop: Good catch — not an intended trade-off. The rationale in the commit message was simply wrong: flake.lock pins the devshell dependencies, so a lock-only bump changing the toolchain under a cached env is a real regression, not an acceptable window. Closing this PR rather than patching it; the two-stat fingerprint is correct as it stands.

kitaebot: Agreed — closing is the right call. The two-stat fingerprint is load-bearing: `flake.lock` pins the toolchain, so a lock-only bump silently invalidates the cached env. One stat on `.envrc` alone can't see that.

amackillop:
Confirmed — closing per the thread. The reviewer's finding stands; the two-stat fingerprint is correct.

Issue #20, opened 2026-08-09
task tool rejects review parameter when LLM passes it as a JSON string


kitaebot:
## Symptom

The `task` tool's `review` parameter rejects calls where the LLM passes it as a JSON string instead of a JSON object. This happened twice in one session:

```
Invalid arguments: invalid type: string "{\"gate\": \"commit\", \"git_ref\": \"...\", \"repo\": \"amackillop/kitaebot\"}", expected struct ReviewMeta
Invalid arguments: invalid type: string "null", expected struct ReviewMeta
```

The workaround is to omit `review` entirely, but that causes a WARN:

```
reviewer call carried no review metadata; recording unlabeled
```

Review findings land in the ledger without their gate/ref/repo metadata, degrading the audit trail.

## Root cause

`src/agent/task.rs` lines 200-226 declare `review` as `Option<ReviewMeta>`:

```rust
#[derive(Deserialize, JsonSchema)]
struct Args {
    prompt: String,
    #[serde(default)]
    agent_type: AgentKind,
    #[serde(default)]
    review: Option<ReviewMeta>,
}

#[derive(Deserialize, JsonSchema)]
struct ReviewMeta {
    repo: String,
    gate: String,
    git_ref: String,
}
```

Deserialization at line 303 uses plain `serde_json::from_value`:

```rust
let args: Args = serde_json::from_value(args)
    .map_err(|e| ToolError::InvalidArguments(e.to_string()))?;
```

When the LLM emits `"review": "{\"repo\":\"...\"}"` (a string containing JSON) instead of `"review": {"repo":"..."}` (an object), serde rejects it with "invalid type: string, expected struct ReviewMeta". There is no fallback to parse the string as JSON.

The LLM's tool-call `arguments` field arrives as a string (`src/types.rs` `ToolFunction.arguments: String`), which `dispatch` (`src/tools/mod.rs` line 288) parses into `serde_json::Value`. At that point a stringified `review` is `Value::String`, not `Value::Object` — and `from_value` into `ReviewMeta` fails.

## Scope

The same class of bug affects every tool with structured (non-string, non-integer) parameters: `github_pr_review_submit` (`event: Event`, `comments: Vec<InlineComment>`), `github_api` (`method: Method`, `body: Option<Value>`), `notify` (`urgency: Urgency`), `review_disposition` (`disposition: Disposition`). The LLM may stringify any of them. `task`'s `review` is the most affected because it's a nested struct, which the LLM is most likely to double-encode.

[...]

PR #22, opened 2026-08-10
Tolerate stringified body in github_api tool

### LINKED ISSUE #10: github_api double-encodes JSON bodies on POST
During the issue #9 plan pass the bot tried to post a comment through the `github_api` escape hatch and hit a 422 from GitHub: "is not an object". Its own diagnosis from the journal (2026-08-06T01:24:29Z):

> github_api tool is broken for POST with body: it double-encodes the JSON body, sending a string instead of an object. GitHub returns 422 "is not an object". Cannot post issue comments via the tool.

Likely the tool serializes the model-provided body argument to a JSON string and then embeds that string as the request body value, rather than passing the object through — so GitHub receives `"{\"body\": ...}"` instead of `{"body": ...}`.

Fix the encoding, and add a test that a POST through `github_api` delivers the payload as an object (the from_fn client fixture can assert on the raw body). Check GET/PUT/PATCH/DELETE paths for the same defect while there.

kitaebot:
The `github_api` tool double-encodes JSON bodies when the LLM passes the `body` parameter as a JSON string (e.g. `"body": "{\"body\": \"done\"}"`) instead of a JSON object. `serde_json::to_vec` on the resulting `Value::String` produces a JSON string `"{\"body\":\"done\"}"` instead of an object `{"body":"done"}`. GitHub rejects this with 422 "is not an object".

Fix: apply the `string_or_value` deserializer to the `body` field. If the value arrives as a string, it parses the inner JSON before deserializing into `Value`, ensuring `to_vec` always sees a `Value::Object`.

Tests assert the raw request body starts with `{` for both stringified and object inputs, and that a GET with no body delivers `None`.

Note: PR #21 (issue #20) adds the same `string_or_value` helper for the `task` tool's `review` field. If both merge, the second will need a trivial conflict resolution (both add the same function to the same location in `tools/mod.rs`).

Closes #10

### DIFF
--- a/specs/03-tools.md
+++ b/specs/03-tools.md
@@ -37,7 +37,8 @@ The LLM sometimes passes a structured field as a JSON string instead of a
 JSON object (e.g. `"review": "{\"repo\":\"...\"}"` instead of
 `"review": {"repo":"..."}`). Fields prone to this use the `string_or_value`
 deserializer, which parses the inner JSON string before deserializing into
-the target type. Applied to `task`'s `review` parameter.
+the target type. Applied to `task`'s `review` parameter and
+`github_api`'s `body` parameter.
 
 ### Per-Turn Context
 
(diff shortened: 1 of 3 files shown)


amackillop:
Rebase on master and fix the conflict

Issue #41, opened 2026-08-13
Devshell evaluation failure is silent; warm runs on bare PATH


amackillop:
## Symptom

The lightning-node-deps duty cloned its repo and warm ran:

```
02:55:12 DEBUG Evaluating direnv dir=.../CumuloGlobal/lightning-node
02:56:42 DEBUG Direnv evaluation complete
02:56:42  WARN warm command failed exit=127 stderr=bash: line 1: just: command not found
```

The 90-second evaluation actually failed: the repo's devshell forces a `git+ssh://` flake input (ln-tool) the VM cannot fetch — reproduced as the daemon user: `error: Failed to fetch git repository 'ssh://git@github.com/CumuloGlobal/ln-tool'`.

## Root cause

`direnv export json` exits 0 even when `.envrc`'s `use flake` fails; it prints the nix error to stderr and exports a base environment. evaluate_direnv (src/tools/direnv.rs) treats exit 0 as success, discards stderr, caches the bare env, and logs "evaluation complete". The failure then surfaces three steps later as `just: command not found` — the real error (an unfetchable SSH input) never appears in any log.

## Proposed fix

Detect the failed-load case and return DirenvError carrying direnv's stderr. Detection options, in preference order: treat the absence of `.envrc`-driven vars (e.g. no `IN_NIX_SHELL` / empty export delta for a `use flake` envrc) as failure; or parse direnv's stderr for its error marker even on exit 0. Either way the warm outcome should read "devshell evaluation failed: <nix error>", not exit-127 from a missing binary.

## Acceptance criteria

- A repo whose devshell cannot evaluate produces a warm/duty outcome naming the underlying nix error (the stderr), not a command-not-found from bare PATH.
- The failed evaluation is not cached (already true today) and the stderr is preserved in the error tee entry.

amackillop:
Live evidence, 2026-08-13 turn 0: mid-turn devshell evals failed on Dependabot branches based before the .envrc split (old `use flake` → ln-tool → SSH), and both exec and git_cli logged 'direnv failed, running without devshell' then proceeded on bare PATH. The turn's builds and hooks degraded quietly, the final git_commit failed downstream of it, and the turn hit max_iterations mid-recovery. The WARN exists but the turn never saw the devshell failure as the blocking fact — the tool result should carry it, not just the journal.

kitaebot:
PR #56 opened: https://github.com/amackillop/kitaebot/pull/56

PR #48, opened 2026-08-13
Cache direnv timeouts with a short TTL to prevent cascade

### LINKED ISSUE #46: Direnv timeout cascade: uncached 900s failures block all operations on a repo
## Symptom

On 2026-08-13, the `lightning-node-dep-queues` duty hit "Maximum iterations reached without completion" after burning its entire iteration budget on direnv timeouts. The timeline:

```
15:33:14 WARN  direnv failed: `direnv export json` timed out after 900s
15:48:24 WARN  direnv failed: `direnv export json` timed out after 900s  (again)
15:57:40 ERROR git_commit exited 1
15:58:04 ERROR Duty error: Maximum iterations reached without completion
```

Two consecutive 900s timeouts = 30 minutes blocked, with every git/exec operation on lightning-node in between also waiting up to 900s each.

## Root cause

`DirenvCache::get()` (`src/tools/direnv.rs:202-204`) does not cache failures:

```rust
Err(e) => {
    // Don't cache failures — next caller retries.
    cache.remove(dir);
    notify.notify_waiters();
    return Err(e);
}
```

This is spec requirement 3 ("Don't cache failures — transient direnv errors don't poison the cache", `specs/03-tools.md:456`). The intent is sound for fast failures (e.g. `.envrc` not allowed, parse error). But `direnv export json` has a 900s timeout (`src/tools/direnv.rs:225`, `timeout_secs: Some(900)`), set high because first-time nix devshell evaluation can legitimately take minutes.

When the nix evaluation itself hangs (e.g. a flake input is unreachable, nix daemon is wedged), each caller retries the full 900s. The failure is not "transient" in the sense the spec assumed — it's a 15-minute blocking operation. Every exec and git operation on that repo during the hang blocks for 900s, cascading into duty timeouts.

## Proposed fix

Cache timeouts (and other slow failures) with a short TTL — e.g. 60s — so repeated operations on a repo with a broken devShell degrade to no-devshell immediately instead of each blocking for 900s. The TTL should be short enough that a genuinely transient failure clears quickly, but long enough to prevent the cascade.

This requires a new `CacheEntry` variant or a `Failed { expires_at: Instant }` entry. The spec's requirement 3 should be amended: "Don't cache failures, except timeouts — those are cached with a short TTL to prevent cascading blocks."

## Evidence

- `src/tools/direnv.rs:225` — 900s timeout
- `src/tools/direnv.rs:204` — `cache.remove(dir)` on failure (no caching)
- `src/tools/direnv.rs:235` — timeout collapsed into `DirenvError::Failed(String)`, losing the structured timeout information
- `specs/03-tools.md:456` — requirement 3 "Don't cache failures"
- Journal: `[2026-08-13T15:58:03Z] [notify] Duty: turn failed: Maximum iterations reached without completion`

kitaebot:
## Summary

When `direnv export json` timed out (900s), the failure was not cached,
so every subsequent git/exec operation on that repo retried the full
900s evaluation. This cascaded into a duty hitting "Maximum iterations
reached without completion" after 30 minutes of blocking (issue #46).

Timeouts are now cached with a 60s TTL so repeated operations during a
hang degrade to no-devshell immediately. Fast failures (blocked, parse
error) remain uncached per the existing spec requirement.

## Changes

- `src/tools/direnv.rs`:
  - `DirenvError::Timeout { secs }` variant to distinguish timeouts
    from other failures (previously collapsed into `Failed(String)`)
  - `CacheEntry::Failed { error, expires_at }` variant to cache
    timeouts with a 60s TTL (`TIMEOUT_TTL`)
  - `get()` checks `Failed` entries in both read and write lock paths;
    expired entries fall through to re-evaluation
  - `evaluate_direnv` maps `ToolError::Timeout` to `DirenvError::Timeout`
  - Eval timeout moved from hardcoded 900s to a `DirenvCache` field
    (test seam: `with_eval_timeout`)
- `specs/03-tools.md`: requirement 3 updated from "Don't cache
  failures" to "Don't cache fast failures" with timeout exception

## Test

`timeout_is_cached_within_ttl`: uses a fake direnv that sleeps 2s
under a 1s eval timeout. First call times out (1s), second call
returns cached error in <500ms. Call-count file proves direnv was
invoked only once.

Closes #46

### DIFF
--- a/specs/03-tools.md
+++ b/specs/03-tools.md
@@ -453,7 +453,12 @@ parallel tool calls each trigger a full `nix print-dev-env` evaluation.
    regardless of concurrent exec calls
 2. **Invalidate on change** — modified `.envrc` or `flake.lock` triggers
    re-evaluation on the next exec call
-3. **Don't cache failures** — transient direnv errors don't poison the cache
+3. **Don't cache fast failures** — transient direnv errors (blocked,
+   parse error) don't poison the cache; the next caller retries.
+   Timeouts are the exception: a 900s evaluation timeout is cached
+   with a short TTL (60s) so repeated operations during a hang degrade
+   to no-devshell immediately instead of each blocking for the full
+   timeout
 4. **Graceful degradation** — if direnv fails, exec runs without the devshell
 5. **Warm on clone** — `git_clone` pre-populates the cache in the background
 6. **Trust before evaluate** — `git_clone` runs `direnv allow` synchronously
--- a/src/tools/direnv.rs
+++ b/src/tools/direnv.rs
@@ -6,8 +6,11 @@
 //! # Cache invalidation
 //!
 //! Two `stat()` calls per lookup: `.envrc` mtime and `flake.lock` mtime.
-//! A changed mtime triggers re-evaluation. Failures are never cached — the
-//! next caller retries.
+//! A changed mtime triggers re-evaluation. Fast failures (blocked, parse
+//! error) are never cached — the next caller retries. Timeouts are cached
+//! with a short TTL so repeated operations during a hang degrade to
+//! no-devshell immediately instead of each blocking for the full 900s
+//! evaluation timeout.
 //!
 //! # Concurrency
 //!
@@ -18,7 +21,7 @@
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
-use std::time::SystemTime;
+use std::time::{Duration, Instant, SystemTime};
 
 use tokio::sync::{Notify, RwLock};
 use tracing::debug;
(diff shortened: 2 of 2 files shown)


amackillop:
Design review: merge-worthy, two annotations.

1. This is deliberately half of the fix and should be read that way: it makes the degradation *immediate*, while #41 remains the other half — making it *visible*. Callers still WARN-and-run-bare-PATH on the cached error; that stays a defect (#41's), just a cheap one now. Do not let this PR's merge close the book on the surfacing work.

2. The `Failed` cache entry checks only `expires_at`, not the fingerprint. The success path re-evaluates when .envrc/flake.lock mtimes change; the failure path should too — store the fingerprint in `Failed` and treat a mismatch as expiry, so fixing the broken envrc is visible on the next call instead of after the TTL. Sixty seconds of serving a stale error to someone who just repaired the cause is small, but the fix is one field.

amackillop (approved):

PR #52, opened 2026-08-13
Gate warm duty on new commits per-repo

### LINKED ISSUE #39: Gate the warm duty on new commits instead of pure schedule
## Motivation

The warm duty re-clones and re-checks every configured repo on a fixed schedule. This week showed both failure modes of pure scheduling: at 24h cadence the first fleet-wide pass OOMed the VM re-verifying nine repos in one window, and the fix — stretching to 7d — bought back the burn by paying in drift-detection latency: a repo whose check breaks on Monday isn't noticed until the following week's pass.

The right trigger is neither. A repo needs re-warming when its remote default branch has moved (new lock, new toolchain, new code) or when its check has never run — not because a week elapsed. The machinery already exists: spec 24's prompt duties support `gate = "new-commits"` via ls-remote with a cursor in the state db that primes silently. The built-in warm duty predates the gate and never adopted it.

## Proposal

Per-repo gating inside the warm duty: on each scheduled tick, ls-remote each configured repo's default branch; warm only repos whose HEAD differs from the stored cursor (or that have no cursor / no checkout). Advance the cursor after a successful warm, so a failed warm retries next tick.

The schedule stays as the tick driver but quiet ticks become one ls-remote per repo — cheap enough that the cadence can tighten back toward daily, recovering the drift-detection latency the 7d stretch gave away, without recovering the burn.

## Notes

- cargo-sweep housekeeping (the `just warm` chain on this repo) wants an occasional run even without commits; either exempt the repo whose check chains sweep, or accept sweep riding only commit-triggered warms.
- Sequentialism and the cgroup memory bounds (vm/configuration.nix) are unchanged by this; the gate reduces how often the heavy path runs, not its shape.

## Acceptance criteria

- A scheduled warm tick against a fleet with no new commits performs no clone, no devshell provision, and no check run — ls-remote per repo only.
- A repo whose remote HEAD moved since its cursor is warmed on the next tick; the cursor advances only on successful warm.
- Duty history records skipped-vs-warmed per repo so the gate's behavior is observable.

kitaebot:
## Summary

The warm duty re-warmed every configured repo on every scheduled tick, whether or not anything had changed. At 24h cadence this wasted cycles on unchanged repos; at higher cadence it risked contention with real work.

Each tick now probes each repo's remote HEAD via `ls-remote` and warms only repos whose HEAD moved past a per-repo cursor (`warm/<nwo>` in the state DB), or that have no cursor (enrollment) or no checkout. The cursor advances only on a successful warm, so a failed warm retries next tick.

## Changes

- **`src/duty/mod.rs`** — New `run_warm()` function replaces the direct `warm_configured_repos()` call. Per-repo `ls-remote` probe, cursor comparison, due-when logic (no cursor, no checkout, or HEAD moved), cursor advance on success only. Six new tests covering skip, warm-on-move, enrollment, missing-checkout, failure-no-advance, and empty-repos.
- **`src/tools/git/git_cli.rs`** — `warm_configured_repos()` replaced by `warm_repos()` (returns keys), `checkout_exists()` (checks `.git`), and public `prepare_and_warm()` (was private). Two existing tests renamed and updated.
- **`specs/03-tools.md`** — Requirement 4 widened from "Unconditional" to "Unconditional per invocation".
- **`specs/24-self-directed-work.md`** — Warm duty description updated with gating; two failure-mode entries added.
- **`README.md`** — Warm duty comment updated.

## Test fixture notes

The test fixture creates a non-bare repo with a commit, then `clone --bare` (avoids cross-device link errors from pushing to an empty bare repo). Sets `commit.gpgsign false` (test process inherits the bot's GPG config), creates a checkout with `.envrc` and a GitHub origin URL (so `origin_nwo` resolves), and leaks the workspace + FakeDirenv tempdir (so they outlive the helper).

Closes #39

### DIFF
--- a/README.md
+++ b/README.md
@@ -216,7 +216,7 @@ kitaebot = {
     };
     duties = {                                   # Duty scheduler (spec 24)
       distill = { every = "1h"; };                # Token gate still applies
-      warm = { every = "24h"; };                  # Build-warm duty; runs only when some repo sets check
+      warm = { every = "24h"; };                  # Build-warm duty; warms repos whose HEAD moved (spec 24)
       self_analysis = {                           # Mine own errors/journal, propose fixes as issues
         every = "24h";
         repo = "owner/repo";                      # Must be trusted and proposal-enabled
--- a/specs/03-tools.md
+++ b/specs/03-tools.md
@@ -496,8 +496,12 @@ visible rather than inside a tool call.
 3. **Two triggers** — the self-maintenance duty
    ([spec 24](24-self-directed-work.md)) for configured repos, and
    checkout preparation for repos cloned on demand
-4. **Unconditional** — a warm-store run costs seconds; no bookkeeping
-   about whether it already happened
+4. **Unconditional per invocation** — a warm-store run itself costs
+   seconds and needs no bookkeeping about whether it already happened
+   (the `Warmer`'s `Notify` dedup handles concurrent callers). The
+   duty's *scheduling* is gated per-repo on new commits
+   ([spec 24](24-self-directed-work.md) warm duty), but each warm
+   invocation that the gate lets through runs unconditionally
 5. **Background** — never blocks the turn that triggered it
 6. **Readers wait, one runner** — a `Notify` per repo, as the direnv
    cache does. Nix serialises same-derivation builds anyway, but a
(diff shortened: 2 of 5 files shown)


amackillop (approved):

amackillop:
Duplicate of #39, which is assigned to you and whose implementation is sitting in your working tree tonight. Your self-analysis dedup only searched bot-authored issues; #39 is human-filed. Dedup-scope defect filed separately.

PR #57, opened 2026-08-14
Enforce output ceiling on GitHub log and API downloads

### LINKED ISSUE #44: job_logs downloads are unbounded; ceiling would keep the wrong end
## Finding

TOOL_OUTPUT_CEILING_BYTES exists to protect the daemon from runaway output filling RAM (its own doc comment), and it is enforced in cli_runner, exec, and grep — but not in the GitHub client. `client.job_logs` returns the raw HTTP body unbounded, and `github_ci_status` concatenates one whole-job log per failed job. A pathological Actions log (or several failed jobs of a large matrix) is exactly the runaway class the ceiling names, arriving over HTTP instead of a pipe.

Context is not at risk — the engine externalizes any tool result over `context.tool_output_tokens`, so oversized logs never enter the live window (surfaced while checking whether CI-log access could blow up duty turns; it can't). This is about the RAM invariant being enforced on some producer paths and not others.

## Second half

`truncate_output` keeps the head and drops the tail. Correct for command output; wrong for CI logs, where the failure concentrates at the end — a head-kept 5 MiB slice of a 20 MiB log is setup noise with the diagnosis amputated. The ceiling application for log downloads should keep the tail (with a '[truncated N leading bytes]' header), or the existing helper gains a direction parameter.

## Acceptance criteria

- `job_logs` (and any other raw-body GitHub download surfaced to tools) applies the ceiling before returning.
- Log-shaped truncation keeps the tail; a truncated log's output states how many leading bytes were dropped.
- A test covers a log exceeding the ceiling: result ends with the log's final lines, not its first ones.

kitaebot:
## Summary

`TOOL_OUTPUT_CEILING_BYTES` protects the daemon from runaway output filling RAM, enforced in `cli_runner`, `exec`, and `grep` — but not in the GitHub client. `job_logs` returned the raw HTTP body unbounded, and `github_ci_status` concatenates one whole-job log per failed job. A pathological Actions log is exactly the runaway the ceiling exists for.

The existing `truncate_output` keeps the head and drops the tail — correct for command output but wrong for CI logs where the failure concentrates at the end.

## Changes

- **`truncate_tail`** (new, `src/tools/mod.rs`) — tail-keeping variant that keeps the last `max_bytes` and prepends `[truncated N leading bytes]`. UTF-8-safe via `ceil_char_boundary`.
- **`job_logs`** (`src/clients/github.rs`) — applies `truncate_tail` with the ceiling before returning (CI logs are tail-weighted).
- **`github_api` raw path** (`src/tools/github/api.rs`) — applies head-keeping `truncate_output` (generic API responses are not tail-weighted).
- **Test helper** (`src/tools/github/test_helpers.rs`) — disables GPG signing on the seed commit so async tests run in environments with `commit.gpgsign` enabled.
- **Spec** (`specs/03-tools.md`) — documents `truncate_tail` and which path uses which truncation direction.

## Test

`oversized_log_keeps_tail` feeds a 2×ceiling log through the full stubbed tool path and asserts the output ends with the diagnosis (tail kept), contains the truncation header, and dropped the head.

Closes #44

### DIFF
--- a/specs/03-tools.md
+++ b/specs/03-tools.md
@@ -92,11 +92,11 @@ hardcoded, so typos are caught by tests instead.
   Context-size limits live in the engines (`context.tool_output_tokens`,
   spec 14), which externalize or truncate tool results long before the
   ceiling matters.
-- **`truncate_tail`** — Tail-keeping variant for log-shaped output where
-  the diagnosis concentrates at the end. Keeps the last `max_bytes` and
-  prepends `[truncated N leading bytes]`. Used by `job_logs` (CI logs);
-  `github_api`'s raw path uses head-keeping `truncate_output` (generic
-  API responses are not tail-weighted).
+- **`truncate_head`** — Head-truncating variant for log-shaped output where
+  the diagnosis concentrates at the end. Drops leading bytes and keeps
+  the last `max_bytes`, prepending `[truncated N leading bytes]`. Used
+  by `job_logs` (CI logs); `github_api`'s raw path uses head-keeping
+  `truncate_output` (generic API responses are not tail-weighted).
 - **`PathGuard`** — workspace-confined path resolution. Rejects null bytes,
   `../`, and absolute paths. Canonicalizes and verifies the result is under the
   workspace root. Provides `resolve()` for existing files and `resolve_new()`
--- a/specs/03-tools.md
+++ b/specs/03-tools.md
@@ -92,6 +92,11 @@ hardcoded, so typos are caught by tests instead.
   Context-size limits live in the engines (`context.tool_output_tokens`,
   spec 14), which externalize or truncate tool results long before the
   ceiling matters.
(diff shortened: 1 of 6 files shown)


specs/03-tools.md
```diff
@@ -92,6 +92,11 @@ hardcoded, so typos are caught by tests instead.
   Context-size limits live in the engines (`context.tool_output_tokens`,
   spec 14), which externalize or truncate tool results long before the
   ceiling matters.
+- **`truncate_tail`** — Tail-keeping variant for log-shaped output where
```
amackillop: `truncate_tail` sounds like it is removing the tail and keeping the head

kitaebot: Renamed to `truncate_head` in bf00ec4. The function truncates the head (drops leading bytes, keeps the tail), so the name now states what it removes, matching `truncate_output`'s convention.

amackillop (approved):

PR #86, opened 2026-08-24
Bound each distillation pass to a token slice

### LINKED ISSUE #47: Distillation diverges when the delta outgrows the iteration budget
## Failure

2026-08-13 17:25: the hourly distill pass died at the sub-agent iteration cap (30) against a ~110k-token delta produced by two failed duty turns. Backlog-carry kept the cursor (correct), but the retry math is divergent: the delta only shrinks on success and grows with every turn of normal operation, so a delta that doesn't fit the fixed budget never distills, and each hourly retry costs real money ($0.46 observed) forever.

## Fix

Bound the work per pass instead of hoping the backlog fits: split the pending delta into slices (per session, or a token ceiling per slice), distill the oldest slice, advance the cursor past that slice on success, repeat next gate-open. Cap failures then affect one slice, the backlog drains monotonically, and a monster day costs several small passes instead of infinite failed big ones.

A per-role iteration budget (memory.distill_max_iterations, defaulting to sub_agents.max_iterations) would also decouple the distiller from explore/worker/reviewer sizing, but chunking is the real fix — no budget survives an unbounded input.

## Workaround in place

deploy config bumps sub_agents.max_iterations 30 → 60 to let today's backlog clear. That bump is a workaround: it raises every sub-agent's worst case, and removing it is part of this issue's acceptance criteria.

## Acceptance criteria

- A delta arbitrarily larger than one pass's budget is fully distilled across successive passes, cursor advancing per slice.
- A slice that fails does not block later slices beyond the next retry of itself.
- The deploy-config max_iterations bump is reverted as part of the change.
- Test: synthetic backlog of N sessions × M tokens distills completely under a budget that fits only one slice.

kitaebot:
Closes #47

## What

Chunked distillation: each pass's consolidated span is bounded by a new `memory.distill_slice_tokens` budget (default unset = the threshold), instead of always gathering a threshold-sized span. A backlog larger than one pass now drains slice by slice across successive gate-open ticks.

## Why

The hourly distill pass died at the sub-agent iteration cap (30) against a ~110k-token backlog, and backlog-carry made that divergent: watermarks only advance on success, so a delta that does not fit one pass retried hourly forever at real cost ($0.46/pass observed). The deploy config's `sub_agents.max_iterations = 60` bump was a workaround; this is the fix, and the bump is reverted here.

## How

- `memory.distill_slice_tokens: Option<u64>` — `None` (default) uses `distill_threshold_tokens`, so default behavior is unchanged. `Some(n)` must be > 0 and <= the threshold; the validation makes `slice > threshold` unrepresentable (that was the bug's exact shape — a 0-sentinel would have allowed it).
- `Distiller` seeds its gather budget from the slice; the gate still keys on the threshold. Gather loop, watermark advance (`after + count`), and failure semantics (watermarks move only on success) are unchanged.
- Deploy config: revert `sub_agents.max_iterations` to the default 30 and set `distill_slice_tokens = 10000` — with the default slice (40k) the reverted cap would recreate the incident configuration.
- Spec 21: execution budget, backlog-carry per-slice drain and retry semantics, config table row, chunking open question resolved. README sample updated.

## Tests

- `oversized_backlog_drains_across_passes`: 3 sessions x 400 tokens against a 400-token slice — one session distilled per pass, gate stays open until drained, no fourth LLM call.
- `failed_slice_retries_then_later_slices_proceed`: a failed pass moves no watermark; the same slice retries, then later slices proceed.
- FakeEngine now honors `after`/`max_tokens` in `transcript_since` and computes pending tokens from transcripts, so the drain tests exercise real watermark arithmetic (matching the clamp-and-progress shape of both real engines).
- Config validation tests: zero slice, slice above threshold rejected; slice at threshold accepted.

Full `just check` (nix flake check, nixfmt/statix/deadnix, cargo deny) passed on both commits; fmt/clippy clean; affected test modules pass (memory 26, config 88, commands 9, agent 99). Full local `cargo test` times out in this environment (known); the nix test check in the gate ran the suite.

## Known residual (pre-existing, out of scope)

A single event larger than the slice still rides the >=1-event progress guarantee into an oversized turn, and a permanently failing head slice blocks later slices until it succeeds — the acceptance criterion's "beyond the next retry" holds for transient failures. Both were true before this change.

### DIFF
--- a/specs/21-memory.md
+++ b/specs/21-memory.md
@@ -132,13 +132,17 @@ reading session history through the context engine abstraction:
   **not** to the head — so history the budget could not reach this tick
   stays pending, not dropped. Between ticks a session can accumulate far
   more than the slice; the gate simply stays open and the next tick
-  folds the next slice. A backlog arbitrarily larger than one pass's
-  budget therefore distills completely across successive passes, the
-  cursor advancing per slice — a pass that fails retries its own slice
-  (watermarks move only on success) and never blocks later slices
-  beyond that retry. Bursty load drains on idle ticks. Sustained load
-  above one slice per tick lags without bound (nothing is lost, memory
-  just trails); draining the backlog within a tick is deferred
+  folds the next slice. Draining stops once the pending total falls
+  below the threshold — with a slice below the threshold, up to
+  `threshold - 1` tokens stay pending until new history reopens the
+  gate; a backlog drains to that floor, not to zero. A pass that fails
+  retries its own slice (watermarks move only on success); a transient
+  failure blocks later slices only until its retry succeeds, while a
+  deterministically failing slice (a fold that exceeds the iteration
+  cap, or poison content) blocks everything behind it — a known
+  residual. Bursty load drains on idle ticks. Sustained load above one
+  slice per tick lags without bound (nothing is lost, memory just
+  trails); draining the backlog within a tick is deferred
   ([FUTURE](FUTURE.md)).
 - **Duties:** extract durable facts from the new events into topics and
   index; merge and dedupe entries the in-turn writes accumulated; prune
@@ -245,7 +249,10 @@ not "no response".
   inside the distiller's window) is provisional; needs live
   token-volume data to confirm.
 - Large consolidated spans could approach the memory model's window at
-  very high thresholds. Resolved by `distill_slice_tokens` (#47): the
-  slice bounds each pass's span, and an oversized backlog drains
-  across successive gate-open passes instead of failing a single
-  oversized turn and retrying forever.
+  very high thresholds. Mitigated by `distill_slice_tokens` (#47): the
+  slice bounds each pass's span, so an oversized backlog drains across
+  successive gate-open passes instead of failing a single oversized
+  turn and retrying forever. The slice bounds spans, not atoms — a
+  single event larger than the slice is still folded whole (the
+  at-least-one-event clamp guarantees progress) and can fail the pass
+  on its own.
--- a/specs/21-memory.md
+++ b/specs/21-memory.md
@@ -113,28 +113,31 @@ reading session history through the context engine abstraction:
   sessions are weighed by their summed `token_count`, and distillation
   runs once that total crosses `distill_threshold_tokens`. Tokens, not
   a raw event count, because the binding constraint is the distiller's
-  context window: the threshold doubles as the shared token budget for
-  the consolidated span, so a triggered pass fits in one turn. Not
-  wall-clock either — an idle week burns nothing, a busy day may cross
-  the gate more than once. The watermark makes the probe a counting
-  query that never loads message bodies.
+  context window. Not wall-clock either — an idle week burns nothing,
+  a busy day may cross the gate more than once. The watermark makes
+  the probe a counting query that never loads message bodies.
 - **Execution:** one **consolidated** pass folds every session's pending
   span into a single distiller turn (cheapest, and it enables
   cross-session dedupe). It runs as a worker sub-agent on a fresh
   ephemeral context, never the root session — reading transcript spans
   is bulk work that must not evict real context. Uses the
   `provider.model_overrides.memory` role when configured. The spans
-  share one token budget seeded from the threshold; each fetch is
-  clamped to the remaining budget and still returns at least one event,
-  so an oversized head cannot stall progress.
+  share one token budget seeded from `distill_slice_tokens` (the
+  threshold when unset); each fetch is clamped to the remaining budget
+  and still returns at least one event, so an oversized head cannot
+  stall progress.
 - **Backlog carry:** exactly one pass runs per heartbeat tick, and it
-  reads at most one budget's worth. Each session's watermark advances by
+  reads at most one slice's worth. Each session's watermark advances by
   the events actually read (positions are dense, so `after + count`),
   **not** to the head — so history the budget could not reach this tick
   stays pending, not dropped. Between ticks a session can accumulate far
-  more than the budget; the gate simply stays open and the next tick
-  folds the next span. Bursty load drains on idle ticks. Sustained load
-  above one budget per tick lags without bound (nothing is lost, memory
+  more than the slice; the gate simply stays open and the next tick
+  folds the next slice. A backlog arbitrarily larger than one pass's
+  budget therefore distills completely across successive passes, the
+  cursor advancing per slice — a pass that fails retries its own slice
+  (watermarks move only on success) and never blocks later slices
+  beyond that retry. Bursty load drains on idle ticks. Sustained load
+  above one slice per tick lags without bound (nothing is lost, memory
   just trails); draining the backlog within a tick is deferred
   ([FUTURE](FUTURE.md)).
 - **Duties:** extract durable facts from the new events into topics and
@@ -222,7 +225,8 @@ not "no response".
 | Config key | Default | Description |
 |------------|---------|-------------|
 | `memory.index_cap_bytes` | 8192 | Injection truncation cap for MEMORY.md |
-| `memory.distill_threshold_tokens` | 40000 (provisional) | Undistilled-token total that opens the distillation gate, and the token budget for the consolidated span |
+| `memory.distill_threshold_tokens` | 40000 (provisional) | Undistilled-token total that opens the distillation gate |
+| `memory.distill_slice_tokens` | unset | Token budget for one pass's consolidated span; unset uses the threshold. Must be > 0 and <= the threshold |
 | `provider.model_overrides.memory` | unset | Model for the distillation pass (falls back to `provider.model`) |
 
 - The index is plain markdown, human-readable and human-editable; no
@@ -241,5 +245,7 @@ not "no response".
   inside the distiller's window) is provisional; needs live
   token-volume data to confirm.
 - Large consolidated spans could approach the memory model's window at
-  very high thresholds. The threshold bounds it in practice, so
-  chunking the span across turns is deferred, not built.
+  very high thresholds. Resolved by `distill_slice_tokens` (#47): the
+  slice bounds each pass's span, and an oversized backlog drains
+  across successive gate-open passes instead of failing a single
+  oversized turn and retrying forever.
--- a/src/config.rs
+++ b/src/config.rs
@@ -370,6 +370,13 @@ pub struct MemoryConfig {
     /// below the distiller's window so a triggered pass fits in one
     /// turn. Must be > 0.
     pub distill_threshold_tokens: u64,
+    /// Token budget for one distillation pass's consolidated span.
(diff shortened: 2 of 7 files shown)


src/memory/distill.rs
```diff
@@ -869,4 +979,168 @@ mod run_tests {
...
+                ("c".into(), session_messages(4, 100)),
+            ]),
+        };
+        let provider = Arc::new(MockProvider::new(vec![
+            Ok(Response::Text("pass 1".into())),
+            Ok(Response::Text("pass 2".into())),
+            Ok(Response::Text("pass 3".into())),
+        ]));
+        let db = crate::state_db::StateDb::open_in_memory().unwrap();
+        let distiller = Distiller::new(&Tools::default(), ws.path(), db, 400, Some(400), 5, 8192);
```
amackillop, line 1007: **Must fix.** Both new tests set threshold 400 / slice Some(400) — numerically identical to the pre-PR budget. Revert the single production line (`Gathered::new(distiller.slice_tokens)` back to `Gathered::new(distiller.threshold)`) and every test in this PR stays green: `Some(400).unwrap_or(400) == 400`, and all pre-existing call sites pass None. The feature line has zero coverage. Add one test with threshold 1200 / slice Some(400) asserting the first pass folds ~400 and later passes drain the rest.

specs/21-memory.md
```diff
@@ -113,28 +113,31 @@ reading session history through the context engine abstraction:
...
-  reads at most one budget's worth. Each session's watermark advances by
+  reads at most one slice's worth. Each session's watermark advances by
   the events actually read (positions are dense, so `after + count`),
   **not** to the head — so history the budget could not reach this tick
   stays pending, not dropped. Between ticks a session can accumulate far
-  more than the budget; the gate simply stays open and the next tick
-  folds the next span. Bursty load drains on idle ticks. Sustained load
-  above one budget per tick lags without bound (nothing is lost, memory
+  more than the slice; the gate simply stays open and the next tick
+  folds the next slice. A backlog arbitrarily larger than one pass's
```
amackillop: **Must fix.** "distills completely across successive passes" is false whenever slice < threshold: the gate still keys on the threshold, so draining stops with up to threshold-1 tokens permanently pending. The motivating shape (110k, threshold 40k, slice 10k) stops at ~30k undistilled until unrelated traffic reopens the gate. State the real bound — a reader debugging missing memory must not chase a phantom slicing bug.

specs/21-memory.md
```diff
@@ -113,28 +113,31 @@ reading session history through the context engine abstraction:
...
+  budget therefore distills completely across successive passes, the
+  cursor advancing per slice — a pass that fails retries its own slice
```
amackillop: **Must fix.** "never blocks later slices beyond that retry" only holds for transient failures. A deterministically failing head slice (fold exceeds max_iterations, poison content) blocks everything: gather restarts at the BTreeMap head, the pass is all-or-nothing before any watermark insert, so no later session ever advances. Your test only covers RateLimited-then-Ok. Weaken the sentence to transient failures, and note the deterministic case as a known residual.

specs/21-memory.md
```diff
@@ -241,5 +245,7 @@ not "no response".
   inside the distiller's window) is provisional; needs live
   token-volume data to confirm.
 - Large consolidated spans could approach the memory model's window at
-  very high thresholds. The threshold bounds it in practice, so
-  chunking the span across turns is deferred, not built.
+  very high thresholds. Resolved by `distill_slice_tokens` (#47): the
+  slice bounds each pass's span, and an oversized backlog drains
```
amackillop: **Must fix.** "Resolved by distill_slice_tokens" overclaims: a single event larger than the slice is still folded whole via the at-least-one-event clamp, fails the same way, and retries every gate-open tick — your own PR body lists this as a residual, and this bullet contradicts it. Reword: the slice bounds spans, not atoms.

src/memory/distill.rs
```diff
@@ -111,6 +111,11 @@ pub struct Distiller {
     system_prompt: String,
     tools: Tools,
     threshold: u64,
+    /// Token budget for one pass's consolidated span: the slice
```
amackillop: Docs: the struct summary above still says the Distiller carries "the gate threshold and the tool-loop cap" — stale now that slice_tokens exists. Also the slice-defaults-to-threshold rationale is stated here, on the constructor, on run(), and in-body: keep one authoritative note on this field, per the terse-comment rule.

src/memory/distill.rs
```diff
@@ -209,8 +220,10 @@ pub async fn run<P: Provider, E: ContextEngine>(
 
     // Share one token budget across sessions so the consolidated span
     // cannot outgrow the distiller's window; each fetch is clamped and
-    // always makes progress.
-    let mut gathered = Gathered::new(distiller.threshold);
+    // always makes progress. The budget is the slice, not the
+    // threshold, so a backlog larger than one pass drains slice by
+    // slice across gate-open ticks.
+    let mut gathered = Gathered::new(distiller.slice_tokens);
```
amackillop, line 220: Non-blocking, file as an issue: slice budget plus fixed BTreeMap order makes session starvation the steady state when one early-alphabet session produces >= slice tokens per tick — later sessions' watermarks never advance while the gate stays open and bills hourly. Pre-PR the budget was 4x the trigger, which masked this. Rotation or largest-backlog-first ordering fixes it.

src/memory/distill.rs
```diff
@@ -496,9 +509,28 @@ mod run_tests {
     impl ContextEngine for FakeEngine {
         async fn pending_distill_tokens(
             &self,
-            _since: &BTreeMap<String, u64>,
+            since: &BTreeMap<String, u64>,
         ) -> Result<BTreeMap<String, u64>, EngineError> {
-            Ok(self.pending.clone())
+            // A static pending map overrides the computed one, so tests
```
amackillop, line 508: Non-blocking: the FakeEngine now hand-rolls a third copy of the clamp/pending arithmetic, and it drifts from the real engines (per-message token flooring vs FlatSession's single division, inverted conversion fallbacks). FlatSession is already an in-memory engine — driving it directly as the double would delete this copy and keep gate-boundary tests honest.

src/config.rs
```diff
@@ -759,6 +766,7 @@ impl Default for MemoryConfig {
         Self {
             index_cap_bytes: 8192,
             distill_threshold_tokens: 40_000,
+            distill_slice_tokens: None,
```
amackillop, line 769: Non-blocking, worth a follow-up: the out-of-box defaults (slice None -> 40k, max_iterations 30) are exactly the #47 incident configuration — only the hand-tuned deploy config is safe, and nothing validates the slice against the iteration budget that actually binds. Consider a bounded default (Some(10_000)) or a validation warning when slice/threshold exceeds what max_iterations plausibly folds.

amackillop (changes requested):
The mechanism is sound and default behavior is genuinely unchanged — the slice/watermark math checks out, and the max_iterations revert is correct per 43ce0e3's acceptance criteria. Gate verified green on this branch locally, and it trial-merges clean against the pending master commits.

Requesting changes for four cheap items, all inline below: one missing test that actually pins the feature, and three spec 21 sentences that promise properties the code doesn't have (specs are contracts here).

Two non-blocking observations that belong in follow-up issues rather than this PR — file them:

1. `/distill` (Gate::Bypass) silently shrank 4x: the manual catch-up now folds one slice (10k deployed) instead of the whole backlog up to threshold, replies as success, and gives no remaining-backlog signal. Either loop Bypass until pending < slice or report remaining tokens in the reply.
2. Slicing multiplies per-pass fixed costs: `transcript_since` has no SQL LIMIT, so each pass materializes every remaining row to keep one slice (~600k tokens read for the motivating 110k backlog), plus per-slice prompt-preamble re-billing. Bound the query.

src/memory/distill.rs
kitaebot: Fixed in ba4694e. Added `slice_below_threshold_bounds_first_pass`: threshold 1200 / slice Some(400), 3 sessions x 400 tokens. Pass 1 folds exactly session "a" (watermark 4, b/c untouched, one provider call); the 800-token remainder closes the gate; adding session "d" (500 tokens, total 1300 >= 1200) reopens it and pass 2 folds "b". Verified the test discriminates: reverting the single production line to `Gathered::new(distiller.threshold)` makes it fail (the threshold budget folds all three sessions in pass 1), restoring it passes.

specs/21-memory.md
kitaebot: Fixed in ba4694e. The backlog-carry section now states the real bound: draining stops once the pending total falls below the threshold — with a slice below the threshold, up to `threshold - 1` tokens stay pending until new history reopens the gate; a backlog drains to that floor, not to zero. The new test pins this too (800-token remainder closes the gate; new history reopens it).

specs/21-memory.md
kitaebot: Fixed in ba4694e. Weakened to transient failures and the deterministic case is now stated as a known residual: "a transient failure blocks later slices only until its retry succeeds, while a deterministically failing slice (a fold that exceeds the iteration cap, or poison content) blocks everything behind it — a known residual."

specs/21-memory.md
kitaebot: Fixed in ba4694e. Reworded to "Mitigated by `distill_slice_tokens`": the slice bounds spans, not atoms — a single event larger than the slice is still folded whole (the at-least-one-event clamp guarantees progress) and can fail the pass on its own.

src/memory/distill.rs
kitaebot: Fixed in ba4694e. Struct summary now lists all three knobs ("the gate threshold, the per-pass slice budget, and the tool-loop cap"). The rationale lives once, on the `slice_tokens` field; the constructor, run(), and gather-site copies are trimmed.

src/memory/distill.rs
kitaebot: Filed as #87 (session starvation under fixed BTreeMap order with a hot early-alphabet session). Agreed it's real but out of scope here — it needs an ordering-policy decision (rotation vs largest-backlog-first), not a one-line fix.

src/memory/distill.rs
kitaebot: Filed as #88 (drive FlatSession directly as the distill test double instead of the hand-rolled FakeEngine arithmetic). Agreed the third copy drifts; replacing it is a test-harness refactor, not a fix-up.

src/config.rs
kitaebot: Filed as #89 (incident-shaped defaults: slice None -> 40k with max_iterations 30; nothing validates the slice against the iteration budget). Deliberately not changing the default in this PR — a bounded default changes out-of-box behavior for every deployment and deserves its own review; the deploy config sets 10000.

amackillop (approved):

PR #94, opened 2026-08-24
Park leftover work before the checkout reset

### LINKED ISSUE #83: Per-turn checkout reset destroys uncommitted work from killed turns
## Symptom

Memory carries a manual ritual that exists only because the per-turn checkout reset destroys work:

> **Turn-boundary checkout resets**: a new GitHub-issue turn can re-clone `projects/<nwo>` at the default branch, wiping uncommitted working-tree changes AND leaving HEAD detached at origin/master. What survives: `.diffs/` files and local branches... Pack the staged diff to `.diffs/` before ending a turn with uncommitted work.

(memory/topics/tooling.md, "Files / exec policy"; indexed in MEMORY.md as "Turn-boundary resets lose uncommitted working trees; `.diffs/` + local branches survive — pack diffs as checkpoints".)

The ritual has already failed once with real cost: issue #47's implementation was lost uncommitted to this reset and had to be mechanically redone. And the ritual is not even a real recovery mechanism: `.diffs/` files are unreadable by exec children (Landlock list-only workspace root), so a packed diff cannot be `git apply`ed — "recovery" means re-issuing every edit from context. The workaround steers every future turn away from ever leaving work uncommitted, which no automatic process will prune.

## Mechanism

`execution_checkout::prepare_at` (src/channel/execution_checkout.rs:46) unconditionally runs `git checkout --force --detach origin/HEAD` then `git clean -fdx` (minus KEPT_CACHES) before **every** execution turn — first turns and resumed turns alike (callers: src/channel/github/issues.rs:161, src/duty/mod.rs:272). `--force` discards tracked-file edits; `clean -fdx` deletes untracked files. A turn that hit the iteration cap, crashed, or simply ended mid-work loses its working tree before any successor turn starts.

The reset's rationale (spec 25, spec 24, module doc) is sound and must be preserved: a turn must start from a current, clean base so a stale clone can never seed an outdated base. The defect is that "clean base" is implemented as "destroy anything present" with no distinction between stale junk and a predecessor's live work.

## Fix direction

Preserve, then reset. Before the destructive step, detect leftover state (dirty tree, or HEAD not at origin/HEAD) and park it where a successor can find it — e.g. an auto-cut branch at the current HEAD with uncommitted changes stashed or committed to it, named deterministically (`kitaebot_recovered/<nwo>` or similar) — then proceed with the reset. The successor turn's ready_note mentions the parked state when present. #62's state reports already tell successors to look for branches; today an uncommitted tree simply has no branch to name.

## Acceptance criteria

- A turn that ends with uncommitted changes in `projects/<nwo>` loses nothing to the next turn's `prepare`: the changes are recoverable (parked branch/stash) without any pre-turn manual action.
- The fresh-base guarantee stands: a clean checkout is still force-detached at origin/HEAD and swept, so a stale clone can never seed an outdated base (existing `prepare_*` tests keep passing).
- **The memory entries documenting the workaround are deleted or rewritten as part of this fix** — the "pack the staged diff to `.diffs/` before ending a turn" ritual in memory/topics/tooling.md and its MEMORY.md index line. The workaround prevents the broken path from ever being re-tested, so no automatic process will prune it; the deletion has to be part of the reviewed work.

amackillop:
execution_checkout::prepare force-detaches at origin/HEAD and cleans before every execution turn, first and resumed alike. The fresh-base guarantee is sound, but "clean base" was implemented as "destroy whatever is present": a turn that died at the iteration cap lost its working tree to the next turn's prepare. Issue #47's implementation went that way and had to be redone from memory, and the bot's memory grew a pack-diffs-to-.diffs ritual that is not even a real recovery path, since exec children cannot read those files back (#83).

Preserve, then reset. On a reused checkout, prepare now parks leftovers on a kitaebot_recovered/<epoch> branch before the destructive steps: a dirty tree (tracked edits plus untracked files) is committed to the branch, and commits stranded on a detached HEAD get a name that survives the re-detach. The successor's ready note names the branch with inspect-and-salvage guidance. Clean checkouts and fresh clones park nothing, and the reset itself is unchanged, so a stale clone still cannot seed an outdated base.

The kept caches are excluded from the parked commit via pathspec, not just from the clean: committing node_modules or .gcroots to the park branch would make the tree-switch back to base delete them from the worktree, silently defeating the cache exclusions (and .gcroots loss would resurrect the issue-75 GC churn).

The park commit pins identity with -c and disables signing: it is a checkpoint, not authored work, and must not depend on the daemon's git config or keyring.

Config::validate moved its model-override checks into a helper; the merged #86 validation pushed the function over the clippy line limit.

The memory entries documenting the .diffs ritual (tooling.md and the MEMORY.md index line) are deleted after this deploys, per the issue's acceptance criteria — deleting them while the old binary still runs would re-expose turns to the destruction the ritual guards against.

Fixes #83

### DIFF
--- a/src/channel/execution_checkout.rs
+++ b/src/channel/execution_checkout.rs
@@ -6,6 +6,14 @@
 //! is dispatched, the checkout is fetched and force-detached at the
 //! remote's default branch, then cleaned, so a stale clone from an
 //! earlier turn can never seed the branch with an outdated base.
+//!
+//! Preserve, then reset: a predecessor turn that died mid-work leaves
+//! a dirty tree or commits stranded on a detached HEAD, and the reset
+//! would destroy them. Leftovers are parked on a `kitaebot_recovered/`
+//! branch first; the ready note names it so the successor can decide
+//! what the work is worth.
+
+use std::path::Path;
 
 use crate::error::ToolError;
 use crate::tools::git::GitCli;
@@ -14,12 +22,34 @@ use crate::tools::git::checkout;
 /// Guidance when no fresh checkout could be prepared for the agent.
 pub(crate) const CLONE_YOURSELF: &str = "Clone or update the repo yourself before branching.";
 
-/// Describe a prepared checkout for the agent.
-pub(crate) fn ready_note(rel: &str) -> String {
-    format!(
-        "A fresh checkout at the default branch is ready at {rel} \
-         (use working_dir: \"{rel}\"). Branch from there; do not clone."
-    )
+/// A checkout [`prepare`] made ready for an execution turn.
+pub(crate) struct Prepared {
+    /// Workspace-relative checkout directory.
+    pub(crate) rel: String,
+    /// Branch holding a predecessor's parked leftovers, when any.
+    pub(crate) parked: Option<String>,
+}
+
+impl Prepared {
+    /// Describe the prepared checkout for the agent.
+    pub(crate) fn ready_note(&self) -> String {
+        use std::fmt::Write as _;
+
+        let mut note = format!(
+            "A fresh checkout at the default branch is ready at {} \
+             (use working_dir: \"{}\"). Branch from there; do not clone.",
+            self.rel, self.rel,
+        );
+        if let Some(parked) = &self.parked {
+            let _ = write!(
+                note,
+                " A previous turn left unfinished work, parked on branch \
+                 {parked}: inspect it before starting (git log/diff \
+                 origin/HEAD..{parked}) and salvage or ignore it."
+            );
+        }
+        note
+    }
 }
 
 /// Workspace-relative checkout directory for `owner/repo`.
@@ -27,24 +57,34 @@ pub(crate) fn checkout_rel_path(nwo: &str) -> Result<String, ToolError> {
     checkout::rel_path("projects", nwo)
 }
 
-/// Clone the repo if needed, fetch, force-detach at the remote's default
-/// branch, and drop any leftover files.
-///
-/// Returns the workspace-relative checkout path.
-pub(crate) async fn prepare(git: &GitCli, nwo: &str) -> Result<String, ToolError> {
+/// Clone the repo if needed, park a predecessor's leftovers, fetch,
+/// force-detach at the remote's default branch, and drop any leftover
+/// files.
+pub(crate) async fn prepare(git: &GitCli, nwo: &str) -> Result<Prepared, ToolError> {
     let rel = checkout_rel_path(nwo)?;
     let url = git.repo_url(nwo);
-    prepare_at(git, &url, &rel).await?;
-    Ok(rel)
+    let parked = prepare_at(git, &url, &rel).await?;
+    Ok(Prepared { rel, parked })
 }
 
 /// In-repo caches the per-turn clean must not touch: re-provisioning
 /// them costs far more than tolerating a stale entry.
 const KEPT_CACHES: &[&str] = &[".direnv", ".gcroots", "node_modules", ".venv"];
 
+/// Branch prefix for parked leftovers. Epoch-suffixed, so repeated
+/// interruptions never collide.
+const RECOVERED_PREFIX: &str = "kitaebot_recovered/";
+
 /// URL-parametrized body of [`prepare`], so tests can use `file://`.
-async fn prepare_at(git: &GitCli, url: &str, rel: &str) -> Result<(), ToolError> {
-    let dir = checkout::ensure_cloned(git, url, rel).await?.into_dir();
+async fn prepare_at(git: &GitCli, url: &str, rel: &str) -> Result<Option<String>, ToolError> {
+    let (dir, parked) = match checkout::ensure_cloned(git, url, rel).await? {
+        // A fresh clone has nothing to lose.
+        checkout::Ensured::Cloned(dir) => (dir, None),
+        checkout::Ensured::Existing(dir) => {
+            let parked = park_leftovers(git, &dir).await?;
+            (dir, parked)
+        }
+    };
     checkout::run(git, &["fetch", "origin"], &dir, true).await?;
     checkout::run(
         git,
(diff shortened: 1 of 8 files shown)


src/channel/execution_checkout.rs
```diff
@@ -62,7 +102,66 @@ async fn prepare_at(git: &GitCli, url: &str, rel: &str) -> Result<(), ToolError>
...
+    let (_, dirty) = checkout::run_read(git, &status, dir).await?;
+    let dirty = !dirty.trim().is_empty();
+    // Exit 0 = HEAD reachable from origin/HEAD = no stranded commits.
+    let (ancestor, _) = checkout::run_read(
+        git,
+        &["merge-base", "--is-ancestor", "HEAD", "origin/HEAD"],
+        dir,
+    )
+    .await?;
+    let stranded = ancestor != 0;
```
kitaebot: **Must fix.** `merge-base --is-ancestor HEAD origin/HEAD` treats any branch tip ahead of origin/HEAD as "stranded". But a *successful* turn ends exactly that way: clean tree, HEAD on the pushed `kitaebot_issue-N` branch tip (the instructed workflow is branch → commit → push → PR, and nothing resets the checkout afterward). So the next `prepare` parks already-pushed work on `kitaebot_recovered/<epoch>` and tells the successor to inspect and salvage it — violating the PR's own "clean checkouts park nothing" invariant and the spec's "stranded detached-HEAD commits" wording. The code never checks whether HEAD is detached.

Consequences: nearly every turn after a successful turn gets a spurious salvage note pointing at already-pushed (possibly already-merged) work, inviting re-do or cherry-pick, plus junk `kitaebot_recovered/*` branch accumulation. The existing `prepare_parks_nothing_on_a_clean_reused_checkout` test only exercises the detached-at-origin/HEAD state that `prepare` itself leaves behind, so it cannot catch this.

Suggested fix — gate the stranded check on HEAD being detached:

```suggestion
    // Only a detached HEAD can strand commits: a branch tip ahead of
    // origin/HEAD is the normal end-state of a successful turn
    // (branch → commit → push → PR), not unfinished work.
    let (attached, _) =
        checkout::run_read(git, &["symbolic-ref", "-q", "HEAD"], dir).await?;
    let stranded = attached != 0 && ancestor != 0;
```

src/config.rs
```diff
@@ -1069,6 +1039,43 @@ impl Config {
     }
 
     /// Validate the channel sections (telegram, github, linear).
+    /// Non-empty override specs, and reasoning bounds (root and
+    /// per-role) that fit the API and leave content room.
+    fn validate_model_overrides(&self) -> Result<(), ConfigError> {
```
kitaebot, line 1043: **Should fix.** The extraction is behavior-preserving, but the doc comment above now documents the wrong function: "Validate the channel sections (telegram, github, linear)." (line 1041) sits on `validate_model_overrides` with the new override doc lines appended beneath it, while `validate_channels` lost its doc entirely. The inserted lines went below the existing doc instead of above the new function.

```suggestion
    /// Non-empty override specs, and reasoning bounds (root and
    /// per-role) that fit the API and leave content room.
    fn validate_model_overrides(&self) -> Result<(), ConfigError> {

kitaebot (commented):
Verdict: **incorrect** — one must-fix, two should-fix.

**Must-fix: the stranded-commit check parks the normal end-state of a successful turn** (inline on src/channel/execution_checkout.rs:132). An execution turn's instructed workflow is "create a branch, implement, commit, push, open a PR" — and nothing resets the checkout afterward. So the common predecessor state at the next `prepare` is: clean tree, HEAD on the *pushed branch tip* of the previous turn. `merge-base --is-ancestor HEAD origin/HEAD` (execution_checkout.rs:126-132) is false for a pushed branch tip, so it's classified as "stranded" and parked — even though the work is already on a named branch and on the remote. That violates the PR's own invariant ("clean checkouts park nothing") and the spec's "stranded detached-HEAD commits" wording, since the code never checks whether HEAD is detached. Consequences: nearly every turn following a successful turn gets a spurious "unfinished work — inspect and salvage" note pointing at already-pushed (possibly already-merged) work, and the repo accumulates junk `kitaebot_recovered/*` branches. The existing clean-checkout test only exercises the detached-at-origin/HEAD state that `prepare` itself leaves behind, so it can't catch this. Fix: gate the stranded check on HEAD actually being detached (`symbolic-ref -q HEAD` failing) or unreachable from any local/remote branch, and add a test for the clean-tree-on-pushed-branch-tip state.

**Should-fix (not anchorable inline — this file is untouched by the PR): the `-c`-prefixed park commit bypasses `hook_timeout`.** `hook_timeout` (src/tools/git/git_cli.rs:288-294) identifies the subcommand via `args.first()`, but the park commit's argv starts with `-c` (execution_checkout.rs:147-153), so `args.first()` is `Some("-c")` and the commit gets the 120s default instead of the 900s timeout that exists precisely for hook-running commits — the park commit is the exact case the long timeout is for. Parse past leading `-c <kv>` pairs when identifying the subcommand. (This is a pre-existing helper meeting a new call site; the fix belongs in this PR since it introduces the call site.)

**Should-fix: the config.rs extraction misplaces a doc comment** (inline on src/config.rs:1044). The refactor is behavior-preserving, but "/// Validate the channel sections (telegram, github, linear)." (config.rs:1041) now documents `validate_model_overrides`, with the new override doc lines appended beneath it, while `validate_channels` lost its doc entirely. The inserted lines went below the existing doc instead of above the new function.

Sound otherwise: the park path is fail-closed (a park failure errors `prepare` before any destructive step, degrading to the `CLONE_YOURSELF` note with the dirty tree surviving for retry); the cache-exclusion pathspec reasoning is correct (`:(exclude)` keeps the caches untracked so the tree-switch cannot delete them, and the test asserts both); the `-c` identity pinning and `commit.gpgsign=false` are correctly placed before the subcommand; the epoch branch name can't realistically collide; and deferring the memory-ritual deletion to post-deploy is the right sequencing.

amackillop:
All three findings addressed in the amended commit (c72e4fb):

**Must-fix — confirmed and fixed.** The stranded check is now gated on HEAD actually being detached (`symbolic-ref -q HEAD` non-zero) before the ancestor probe runs, so the normal post-turn state (clean tree, HEAD on the pushed branch tip) parks nothing. Added the test you asked for: `prepare_parks_nothing_after_a_successful_turn_on_a_branch` builds exactly that state and asserts no park, no recovered branches, and the work branch untouched. Good catch — the existing clean-checkout test only covered the detached-at-origin/HEAD state prepare itself leaves, which is why it couldn't see this.

**hook_timeout — fixed, plus a stronger fix at the call site.** `hook_timeout` now parses past leading `-c key=val` pairs (with a test case). But the park commit also gained `--no-verify`, which supersedes the timeout concern for this call site and fixes something your finding pointed at indirectly: a pre-commit hook that gates on the tree state (this repo's own hook runs the full gate) would fail against exactly the broken tree the park exists to rescue, aborting prepare. A checkpoint of possibly-broken work must not be hook-gated at all.

**Doc placement — fixed.** `validate_channels` has its doc back; the overrides doc sits on `validate_model_overrides` alone.

kitaebot (approved):
All three findings verified fixed in c72e4fb, no new issues in the delta.

- **Stranded check (must-fix):** `symbolic-ref -q HEAD` is the right probe and runs before the ancestor probe — attached HEAD short-circuits to not-stranded. The new test `prepare_parks_nothing_after_a_successful_turn_on_a_branch` constructs exactly the misparked state (clean tree, attached HEAD one commit ahead of origin/HEAD) and asserts no park, no recovered branches, work branch untouched; it would have failed against the old logic. The gate opens no hole: detached-at-origin/HEAD still parks nothing via the ancestor probe, and an attached-but-dirty tree still parks via the independent dirty check.
- **`--no-verify`:** sound, and the right call — the pre-commit hook runs the full gate against exactly the broken tree the park exists to rescue. Composes correctly with the `-c` identity pinning and `commit.gpgsign=false` (orthogonal mechanisms). The `hook_timeout` fix is still correct to have: the `while let ["-c", _, tail @ ..]` strip is right and tested, and it covers any future `-c`-prefixed hook-running call site.
- **Doc placement:** restored correctly on both functions.

The three `prepare` callers, the `Prepared` struct plumbing, and the spec 25 wording all check out against the implemented detached-only semantics.

PR #95, opened 2026-08-25
Fold own-PR feedback into one turn per PR per tick

### LINKED ISSUE #92: GitHub feedback pass dispatches one agent turn per review comment
## Evidence

2026-08-24, PR #86: a human review (CHANGES_REQUESTED) carried a verdict plus 8 inline diff comments, submitted as one review event. The GitHub channel dispatched 9 sequential agent turns for it: the verdict turn processed all findings (fixes in ba4694e, every inline thread replied), and then 8 redundant per-comment turns followed, one per completed turn — each a root-model turn re-delivering a byte-identical comment already answered in its thread. Two observed so far; the remainder are queued behind the actor's serial mailbox and will arrive as further no-op turns.

Not a re-dispatch bug: `last_poll` advances past the comments' `created_at` after the poll, and each comment is dispatched exactly once. The defect is granularity.

## Cost

- N inline comments on one review = N+1 agent turns where 1 suffices. Each is a full root-model turn, with session history growing per turn.
- The actor consumes envelopes serially and `feedback_pass` awaits each `send()` in turn, so the dribble blocks Telegram, duties, and all other GitHub work for the full duration — hours at root-model latency.
- Each redundant turn is saved from re-doing already-done work only by noticing the session history.

## Code

`src/channel/github/prs.rs`, `feedback_pass` (~206-320): three loops — reviews, conversation comments, diff comments — each calling `send()` per item, i.e. one agent turn per item. `send()` (~550) awaits the full turn.

## Spec

Spec 20 batches the other feedback passes: pass 3 "both in one tick fold into a single combined turn", pass 4 "new trusted feedback folds into one discussion turn per PR". Pass 1's granularity is unstated, but per-item fan-out contradicts the batching design those passes follow — a review is one logical event.

## Fix direction

Batch all new feedback for one PR in one tick — verdicts, conversation comments, and diff comments — into a single dispatch message, matching passes 3/4. The processing turn already fetches the full comment set via `github_pr_diff_comments` when it needs detail, so the batched message can stay compact.

kitaebot:
Closes #92

## Problem

`feedback_pass` dispatched one agent turn per review, per conversation comment, and per diff comment on the bot's own PRs. A review event with N inline comments produced N+1 sequential root-model turns where 1 suffices (observed on PR #86: 9 turns, 8 of them no-ops re-delivering comments the verdict turn had already answered). The actor consumes envelopes serially, so the dribble blocked Telegram, duties, and all other GitHub work for hours at root-model latency. Not a re-dispatch bug — `last_poll` advances past every item after the poll — the defect was granularity, contradicting the batching spec 20 already mandates for the tracked and contributed passes.

## Change

Two commits:

1. **463cedb — Restructure the feedback pass into a pure decision core.** Behavior-preserving: `feedback_pass` becomes fetch → `decide_feedback` (pure) → send-loop, the shape the tracked and contributed passes already use. Filters, formats, per-item granularity, and count semantics unchanged. One error-path improvement: sends now run only after all fetches succeed, so a mid-pass fetch failure dispatches nothing instead of leaving earlier PRs' items dispatched-but-unrecorded (which the next tick would duplicate, since `last_poll` doesn't advance on error).

2. **9a637f7 — Fold own-PR feedback into one turn per PR per tick.** All new feedback on one PR (reviews, conversation comments, diff comments) folds into a single turn message per tick, matching passes 3/4: replies must not race each other on the same branch, and a review is one logical event. The diff-comment item format gains the comment id so the turn can reply in-thread via `github_pr_diff_reply` without a fetch round-trip — same reason the contributed pass's inline format carries it. Spec 20 updated (batching rule + id in the format bullet).

Filters are untouched throughout: not the bot's own, pending reviews invisible, newer than `last_poll`, trusted author (with warn), bodyless approvals dropped (with debug).

## Tests

Four new unit tests over the pure core: folding all three feedback kinds into one turn, per-PR (not cross-PR) grouping, the skip filters, and the bodyless-approval skip; the two `format_diff_comment` tests updated for the id. Full `just check` gate green on both commits (clippy, fmt, tests, nix flake check).

## Manual verification

The PR #86 scenario is the test case: a review with a verdict plus 8 inline comments now yields one dispatch (one turn message containing 9 item blocks), not 9 turns. The batched message's closing instruction references the Developer Workflow's review-feedback step, which author-role turns carry as their prompt segment.

### DIFF
--- a/specs/20-github.md
+++ b/specs/20-github.md
@@ -215,9 +215,11 @@ formats for the items inside the batched message:
 - Diff comment: `Inline comment on PR #5 "Title" (owner/repo) by @dave at src/main.rs:42 (comment id 12):\n\nBody`
 
 The diff comment carries its id so the turn can reply in-thread
-(`github_pr_diff_reply`) without a fetch round-trip. The batched
-message stays compact: the turn fetches the full comment set via
-`github_pr_diff_comments` when it needs detail.
+(`github_pr_diff_reply`) without a fetch round-trip. Item bodies are
+inlined verbatim and unbounded — they are the work order, and a
+truncated body is a dropped request. The fold bounds turn count, not
+message size; oversized messages are externalized downstream by the
+context engine ([spec 14](14-context-engine.md)).
 
 ### Review requests
 
--- a/specs/20-github.md
+++ b/specs/20-github.md
@@ -187,11 +187,20 @@ For each of the bot's open PRs, fetch reviews
 (`/repos/{nwo}/pulls/{n}/reviews`), conversation comments
 (`/repos/{nwo}/issues/{n}/comments`), and inline diff comments
 (`/repos/{nwo}/pulls/{n}/comments`). Skip the bot's own items,
-items older than `last_poll`, and untrusted authors. Message formats:
+items older than `last_poll`, and untrusted authors. All new
+feedback on one PR folds into **one turn per PR per tick** — replies
+must not race each other on the same branch, and a review with N
+inline comments is one logical event, not N+1 turns. Message formats
+for the items inside the batched message:
 
 - Review: `Review on PR #5 "Title" (owner/repo) by @alice: APPROVED\n\nBody`
 - Comment: `Comment on PR #5 "Title" (owner/repo) by @carol:\n\nBody`
-- Diff comment: `Inline comment on PR #5 "Title" (owner/repo) by @dave at src/main.rs:42:\n\nBody`
+- Diff comment: `Inline comment on PR #5 "Title" (owner/repo) by @dave at src/main.rs:42 (comment id 12):\n\nBody`
+
+The diff comment carries its id so the turn can reply in-thread
+(`github_pr_diff_reply`) without a fetch round-trip. The batched
+message stays compact: the turn fetches the full comment set via
+`github_pr_diff_comments` when it needs detail.
 
 ### Review requests
 
--- a/specs/20-github.md
+++ b/specs/20-github.md
@@ -37,10 +37,11 @@ four passes:
 2. **Review requests**: `GET /search/issues` with
    `is:pr is:open review-requested:{bot}`.
 3. **Tracked reviewed PRs**: for each PR in the `reviewed`
-   map, fetch state, head SHA, and new comments. Closed/merged PRs are
-   pruned; a new head SHA triggers an incremental re-review; new trusted
-   comments trigger a discussion turn; both in one tick fold into a
-   single combined turn.
+   map, fetch state, head SHA, comments, and reviews (the reviews
+   date review-linked diff comments by `submitted_at`). Closed/merged
+   PRs are pruned; a new head SHA triggers an incremental re-review;
+   new trusted comments trigger a discussion turn; both in one tick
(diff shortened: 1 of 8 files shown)


amackillop (changes requested):
The fold design and the two-commit split are right, and commit 1 standing alone was verified locally. Two error-path findings, both small:

1. **Per-PR fetch failure now starves everything** (feedback_pass keeps `?` on fetch_pr_feedback/pull_comments). One persistently broken PR (repo gone private, 404 every tick) aborts the pass before any dispatch, and since last_poll doesn't advance on error, every other own PR's feedback is dead until the broken PR closes. contributed_pass already settled this tradeoff the other way — 'one broken PR must not wedge the cursor' (line 508) — skip-and-log the broken PR, matching that precedent. Your all-fetches-first rationale (no dispatched-but-unrecorded duplicates) still holds for the PRs that fetched clean.

2. **The snapshot-all-then-dispatch shape widens the lost-comment window.** Feedback arriving on PR B while PR A's folded turn runs postdates B's snapshot but predates the post-turn `last_poll = now_iso8601()` (line 183), so next tick it's filtered as old and silently dropped. Pre-existing in kind (the old interleaving had this for the last PR only), but now every PR shares the earliest snapshot. Bound last_poll by the fetch time instead of post-dispatch now, and the window closes for all passes.

Nit, no change required: the 'dispatched N items' log now counts folded turns, not items — fine, but worth a word in the log key if you touch that line anyway.

amackillop:
Two additions from a deeper pass, one significant:

3. **A pending review's inline comments can be filtered out of the very event they belong to.** Diff comments gate on `created_at`, reviews on `submitted_at` — and GitHub stamps a pending-review comment's `created_at` at *draft* time. Reviewer drafts 5 comments at 11:50, a tick advances last_poll at 12:00, review submits at 12:10: the review passes the filter, all 5 comments are older than last_poll, and the folded event dispatches with 0 of its N comments — permanently, since they stay older than every future last_poll. Pre-existing filter, but this PR's one-event contract makes it load-bearing: gate a review's diff comments on their review's `submitted_at` (or fetch them by review id) rather than their own `created_at`. Draft-then-submit is the normal way multi-comment reviews get written, so this is the common case, not an edge.

4. Smaller, spec honesty: the folded message inlines every item body verbatim and unbounded, while the new spec 20 text claims 'the batched message stays compact'. LCM externalization catches truly huge messages downstream, but the spec and code currently disagree — either bound/summarize bodies past a size and point at github_pr_diff_comments for detail, or drop the compactness claim.

amackillop (approved):
All four findings addressed, verified against the delta: a5204fc (skip-and-log matches the contributed_pass precedent), 1ec062c (last_poll pinned to tick start closes the lost-comment window), 7ff5e9c (review-linked diff comments dated by their review's submitted_at, applied to all three item filters, with fixture-backed e2e coverage of the draft-then-submit case), a33c90e (spec 20 states the real size contract). The .diffs/ gate-artifact disclosure is acceptable for these commits since staged diff equaled tree; the underlying sandbox friction is a known class. Ready to merge — squash recommended, the fix commits served their review purpose.

PR #106, opened 2026-08-27
file_edit: hard-stop repeated futile payloads

### LINKED ISSUE #102: file_edit: return a post-edit preview and hard-stop repeated no-op edits
## Problem

Two retry-cycle sinks in `src/tools/file_edit.rs`:

1. After a successful edit the model typically re-reads the file to confirm the result and anchor the next edit — a full Read round-trip per edit in multi-edit sequences, billed every time.
2. A no-op or repeatedly failing edit payload has no backstop short of the iteration cap. Nothing stops the loop from re-submitting the same payload many times.

## Proposal

- **Close the loop in the result.** On success, append a numbered excerpt of the edited region (post-edit line numbers, a few lines of context) to the tool result. The next edit in the same region anchors without a re-read.
- **No-op loop guard.** Track recent edit payloads per path. The third identical payload that produced no change (or the identical failure) returns a hard error instructing a re-read — a distinct variant from the existing ambiguity and stale-read errors, per the error-ADT rules.

Both are pure additions next to the existing rung-ladder logic, unit-testable alongside its tests.

amackillop:
Closes #102.

The issue's first half (post-edit preview) already landed in b70a861: every success echoes the edited region with line numbers and three lines of context, and spec 03 documents it under "Success result". This PR implements the remaining half, the no-op loop guard, plus one upfront rejection the guard made obvious.

**Reject `old_string == new_string`.** A payload with identical old and new strings is a no-op by definition, but the tool accepted it, rewrote the file with the same bytes, and returned a normal success echo. It now fails upfront as `InvalidArguments`.

**Hard-stop repeated futile payloads.** The tool remembers recent futile payloads per path as u64 fingerprints in a short ring. A payload is futile when it fails to match (or matches ambiguously) or when it matches but leaves the file byte-identical, which a fuzzy rung can do even with differing strings. The third identical payload in the window returns `EditLoop`, a new `ToolError` variant instructing a re-read, instead of a third copy of the same result. No-change successes are labeled `(no change)` and skip the write; a content-changing edit clears the path's history. The guard classifies outcomes and never pre-blocks: if the file changed since the failures and the same payload now lands, it just succeeds.

Spec 03 gains a "No-op loop guard" section and an error-mode table row. `error_class` in the agent loop groups the new variant as `edit_loop` for strike escalation.

### DIFF
--- a/src/agent/mod.rs
+++ b/src/agent/mod.rs
@@ -1808,16 +1808,16 @@ mod tests {
                 Some(FailureKind::Blocked),
             ),
             (
-                Err(ToolError::Precondition("exit 1".into())),
-                Some(FailureKind::Precondition),
-            ),
-            (
                 Err(ToolError::EditLoop {
                     path: "src/main.rs".into(),
                     attempts: 3,
                     outcome: crate::error::EditFutility::NoChange,
                 }),
                 Some(FailureKind::EditLoop),
+            ),
+            (
+                Err(ToolError::Precondition("exit 1".into())),
+                Some(FailureKind::Precondition),
             ),
             (
                 Err(ToolError::InvalidArguments("missing field".into())),
--- a/src/agent/mod.rs
+++ b/src/agent/mod.rs
@@ -1812,6 +1812,14 @@ mod tests {
                 Some(FailureKind::Precondition),
             ),
             (
+                Err(ToolError::EditLoop {
+                    path: "src/main.rs".into(),
+                    attempts: 3,
+                    outcome: crate::error::EditFutility::NoChange,
+                }),
+                Some(FailureKind::EditLoop),
+            ),
+            (
                 Err(ToolError::InvalidArguments("missing field".into())),
                 Some(FailureKind::InvalidArguments),
             ),
--- a/src/agent/mod.rs
+++ b/src/agent/mod.rs
@@ -907,6 +907,7 @@ fn error_class(err: &ToolError) -> String {
         ToolError::Blocked { .. } => "blocked".into(),
         ToolError::Cancelled => "cancelled".into(),
         ToolError::CommandFailed { exit_code, .. } => format!("command_failed:{exit_code}"),
+        ToolError::EditLoop { .. } => "edit_loop".into(),
         ToolError::FtsQuery { .. } => "fts_query".into(),
         ToolError::Github(e) => match e {
             crate::error::GithubError::Api { status, .. } => format!("github:api:{status}"),
--- a/src/error.rs
+++ b/src/error.rs
@@ -169,6 +169,24 @@ impl ProviderError {
 #[error("tool name {0:?} must match ^[a-zA-Z0-9_-]+$")]
 pub struct InvalidToolName(pub String);
 
+/// What every attempt in a [`ToolError::EditLoop`] produced.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum EditFutility {
+    /// Each attempt returned the same match failure.
+    FailedIdentically,
+    /// Each attempt matched but left the file byte-identical.
+    NoChange,
+}
+
+impl std::fmt::Display for EditFutility {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::FailedIdentically => f.write_str("failed identically"),
+            Self::NoChange => f.write_str("produced no change"),
+        }
+    }
+}
+
 /// Tool execution errors.
 #[derive(Debug, Error)]
 pub enum ToolError {
@@ -203,6 +221,23 @@ pub enum ToolError {
         exit_code: i32,
         /// `$ command`, stdout, stderr, `Exit code: N`.
         output: String,
+    },
+
+    /// The same `file_edit` payload was futile several times running:
+    /// the model is looping on a stale view of the file. No `#[source]`
+    /// carrying the repeated result — the model already received it
+    /// verbatim twice; the payload here is the instruction to stop.
+    #[error(
+        "file_edit on {path} {outcome} {attempts} times running; \
+         re-read the file to re-anchor before editing it again"
+    )]
+    EditLoop {
+        /// The workspace-relative path the edits targeted.
+        path: String,
+        /// Occurrences of this payload in the path's recent window.
+        attempts: usize,
+        /// What every attempt produced.
+        outcome: EditFutility,
     },
 
     /// FTS5 rejected an `lcm_grep` search pattern as query syntax and
@@ -429,6 +464,7 @@ impl ToolError {
             // what the service returns, so they log whole.
             Self::Blocked { .. }
             | Self::Cancelled
+            | Self::EditLoop { .. }
             | Self::FtsQuery { .. }
             | Self::Github(_)
             | Self::Http { .. }
--- a/src/tools/file_edit.rs
+++ b/src/tools/file_edit.rs
@@ -6,9 +6,11 @@
 //! first rung with at least one match decides the outcome.
 
 use std::cmp::Reverse;
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet, VecDeque};
 use std::future::Future;
+use std::hash::{DefaultHasher, Hash, Hasher};
 use std::pin::Pin;
+use std::sync::Mutex;
 
 use schemars::JsonSchema;
 use serde::Deserialize;
@@ -16,7 +18,7 @@ use tracing::debug;
 
 use super::path::PathGuard;
 use super::{Tool, ToolCtx, truncate_output};
-use crate::error::ToolError;
+use crate::error::{EditFutility, ToolError};
 
 #[derive(Deserialize, JsonSchema)]
 struct Args {
@@ -37,11 +39,38 @@ struct Args {
 /// Tool that performs find-and-replace edits on workspace files.
 pub struct FileEdit {
     guard: PathGuard,
+    history: Mutex<EditHistory>,
 }
 
 impl FileEdit {
     pub fn new(guard: PathGuard) -> Self {
-        Self { guard }
+        Self {
+            guard,
+            history: Mutex::new(EditHistory::default()),
+        }
+    }
+
+    /// Record a futile attempt; the third identical payload in the
+    /// path's recent window becomes a hard stop.
+    fn record_futile(
+        &self,
+        path: &str,
+        fingerprint: u64,
+        outcome: EditFutility,
+    ) -> Result<(), ToolError> {
+        let attempts = self
+            .history
+            .lock()
+            .expect("edit history mutex poisoned")
+            .record_futile(path, fingerprint);
+        if attempts >= FUTILE_LIMIT {
+            return Err(ToolError::EditLoop {
+                path: path.to_string(),
+                attempts,
+                outcome,
+            });
+        }
+        Ok(())
     }
 }
 
@@ -89,7 +118,9 @@ impl Tool for FileEdit {
                 source: e,
             })?;
 
+            let fingerprint = fingerprint(&args);
             let Some((rung, spans)) = find_matches(&content, &args.old_string) else {
+                self.record_futile(&args.path, fingerprint, EditFutility::FailedIdentically)?;
                 return Err(ToolError::Precondition(stale_read_message(
                     &content,
                     &args.old_string,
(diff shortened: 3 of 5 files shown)


src/tools/file_edit.rs
```diff
@@ -128,6 +176,53 @@ impl Tool for FileEdit {
...
+/// consecutive counter, so interleaving two failing payloads still
+/// trips the guard for each.
+#[derive(Default)]
+struct EditHistory(HashMap<String, VecDeque<u64>>);
+
+impl EditHistory {
+    /// Record a futile attempt; returns how many times this
+    /// fingerprint appears in the path's recent window, this one
+    /// included.
+    fn record_futile(&mut self, path: &str, fingerprint: u64) -> usize {
```
kitaebot, line 205: The guard's "identical failure keyed on payload alone" argument holds only under a single writer per path. The agent loop executes parallel tool calls via `join_all`, and `task(worker)` sub-agents share the same workspace and the same `Arc`ed `FileEdit` instance — two workers editing one file can race: a successful write from one clears the ring the other was accumulating, or interleaved workers each trip `EditLoop` on payloads that would have succeeded. Either state the single-writer assumption in spec 03's "No-op loop guard" section and the commit message, or bound the cross-agent risk explicitly (e.g. key the history by path only, acknowledge races are benign because the guard never pre-blocks and a spurious `EditLoop` costs one re-read).

src/tools/file_edit.rs
```diff
@@ -128,6 +176,53 @@ impl Tool for FileEdit {
...
+        if self.0.len() >= MAX_PATHS && !self.0.contains_key(path) {
+            self.0.clear();
```
kitaebot:
```suggestion
    fn record_futile(&mut self, path: &str, fingerprint: u64) -> usize {
        if self.0.len() >= MAX_PATHS && !self.0.contains_key(path) {
            self.0.clear();
        }
```
This global `clear()` drops every path's history at once — including a path one attempt from its hard stop — and on a daemon-lifetime tool over a workspace with many checkouts, 32 paths is the realistic overflow path. It silently disarms the guard globally exactly when it matters. Evict a single entry instead (LRU or FIFO on first-insert order; a `VecDeque<String>` of insertion order alongside the map is enough) so only the oldest path restarts. The `history_path_cap_resets` test should then assert the evicted path restarts at one while a recent path keeps its count.

kitaebot (commented):
Verdict: **incorrect** — two must-fix gaps in the guard's semantics, one observability should-fix. The core mechanics (ring of 8 fingerprints per path, occurrence-count threshold of 3, clearing on content change, upfront identity rejection, `EditLoop` ADT shape) are sound and well-tested; the problems are in what the guard does when its assumptions break.

**Must-fix 1 — unstated single-writer assumption.** The guard's safety argument ("identical failure keyed on payload alone is sound because the ladder is deterministic on unchanged content") holds only with one writer per path. But the agent loop runs parallel tool calls via `join_all`, and `task(worker)` sub-agents share the same workspace and the same `Arc`ed `FileEdit` instance. Two workers on one file can race: one worker's successful write clears the ring the other was accumulating, or interleaved workers each trip `EditLoop` on payloads that would have succeeded. Either state the serialization assumption in spec 03 and the commit message, or bound the risk explicitly.

**Must-fix 2 — `MAX_PATHS` overflow is a global wipe.** At 32 tracked paths, one new path fires `self.0.clear()`, dropping every path's history — including a path one attempt from its hard stop. On a daemon-lifetime tool over a workspace with many checkouts this is the realistic overflow path, and it silently disarms the guard globally exactly when it matters. Evict a single entry (LRU) instead of clearing the map; the `history_path_cap_resets` test then asserts only the evicted path restarts.

**Should-fix — `EditLoop` is invisible to `classify_failure`.** The Display text (`file_edit on {path} …`) matches no prefix the failure classifier knows, so it lands in `FailureKind::Other` in both the `/stats` tool-errors table and the error tee. The PR wires `error_class = "edit_loop"` for the strike tracker but leaves the observability layer unable to count the new class by name — the whole point of the guard is observing this pathology. Add a `classify_failure` arm (or reword the message to a recognized shape).

One nit on the `record_futile` doc comment (occurrences-in-window, not consecutive — one clause would save the next reader the ring walk). The no-`#[source]` decision for `EditLoop` is sound: the source would be the prior `Precondition` payload the model already saw twice, and carrying it would re-embed the excerpt the log cap exists to avoid.

Findings ledger: 115 (concurrency assumption), 116 (MAX_PATHS wipe), 117 (classify_failure blind spot), 118 (doc nit).

src/error.rs
```diff
@@ -205,6 +223,23 @@ pub enum ToolError {
         output: String,
     },
 
+    /// The same `file_edit` payload was futile several times running:
+    /// the model is looping on a stale view of the file. No `#[source]`
+    /// carrying the repeated result — the model already received it
+    /// verbatim twice; the payload here is the instruction to stop.
+    #[error(
+        "file_edit on {path} {outcome} {attempts} times running; \
```
kitaebot, line 231: `EditLoop`'s Display text (`file_edit on {path} … times running; re-read the file …`) matches no `classify_failure` prefix, so it lands in `FailureKind::Other` in both the `/stats` tool-errors table and the error tee. The PR wires `error_class = "edit_loop"` for the strike tracker but leaves the observability layer unable to count the new class by name — and observing this pathology is the point of the guard. Add a `classify_failure` arm for the `Error: file_edit on ` prefix (or reword the Display message to a shape the classifier already recognizes).

src/tools/file_edit.rs
```diff
@@ -128,6 +176,53 @@ impl Tool for FileEdit {
...
+
+/// Recent futile edit payloads per path — attempts that failed to
+/// match or matched without changing the file. A ring rather than a
+/// consecutive counter, so interleaving two failing payloads still
+/// trips the guard for each.
+#[derive(Default)]
+struct EditHistory(HashMap<String, VecDeque<u64>>);
+
+impl EditHistory {
+    /// Record a futile attempt; returns how many times this
```
kitaebot: One clause would save the next reader the ring walk: the count is occurrences-in-window, not consecutive — the doc currently reads as consecutive unless you already know the ring semantics.

src/tools/file_edit.rs
amackillop: Documented in 919ffdd: spec 03 now states the history is shared across tasks and sub-agents, names both races, and records why they are benign (a cleared ring only delays the hard stop; a spurious EditLoop costs one re-read). Chose documentation over serialization since the guard needs no lock to stay correct.

src/tools/file_edit.rs
amackillop: Fixed in 72511d4: first-record order tracked in a VecDeque beside the map; overflow evicts only the front entry, and clear() untracks the path so a relapsing path is not double-tracked. Test renamed to history_path_cap_evicts_only_the_oldest_path and asserts the survivor keeps its count.

src/error.rs
amackillop: Fixed in bbf5325: FailureKind::EditLoop with a classify_failure arm on the "Error: file_edit on " prefix, pinned in the producer/consumer round-trip test in agent/mod.rs.

src/tools/file_edit.rs
amackillop: Fixed in 72511d4 (folded into the eviction commit, same function): the doc now reads "total occurrences anywhere in the path's recent window (not a consecutive run)".

src/tools/file_edit.rs
```diff
@@ -665,13 +772,169 @@ mod tests {
...
+    #[test]
+    fn history_path_cap_evicts_only_the_oldest_path() {
+        let mut history = EditHistory::default();
+        for p in 0..MAX_PATHS {
+            history.record_futile(&format!("p{p}"), 7);
+        }
+        assert_eq!(history.record_futile("overflow", 7), 1);
+        // p0 was evicted and restarts; p1 kept its count.
+        assert_eq!(history.record_futile("p1", 7), 2);
+        assert_eq!(history.record_futile("p0", 7), 1);
```
kitaebot, line 924: The test's final assertion (`record_futile("p0", 7)` at line 923) re-tracks the evicted p0 while the map is still at `MAX_PATHS`, which silently evicts p1 — the survivor whose retention the test just pinned. Harmless to the assertion's outcome, but the cascading eviction is invisible to a reader and the re-track-while-full path gets no dedicated coverage; consider clearing a path first or asserting the second eviction explicitly.

src/agent/mod.rs
```diff
@@ -1810,6 +1811,14 @@ mod tests {
                 Err(ToolError::Precondition("exit 1".into())),
                 Some(FailureKind::Precondition),
             ),
+            (
```
kitaebot, line 1810: The EditLoop round-trip case sits between Precondition and InvalidArguments; the same diff keeps the `FailureKind` variants, Display arms, and `classify_failure` arms alphabetical, so slotting this case between CommandFailed and Precondition would keep the table consistent with its siblings.

kitaebot (approved):
Verdict: **correct** — the delta resolves all four prior findings, no new defects.

- **Concurrency story (919ffdd):** spec 03 now states the history is one map shared across tasks and sub-agents, names both races, and the benignness argument matches the code — the guard records only on failure paths and clears before a successful write, so it never pre-blocks; a spurious `EditLoop` costs exactly one re-read, which the error text itself instructs. Documentation over serialization is the right call for a heuristic guard.
- **Overflow eviction (72511d4):** first-record order in a `VecDeque` beside the map; only the front entry is evicted at capacity; `clear()` untracks the path so relapses can't be double-tracked. Order consistency holds under the mutex, and the test pins the survivor's count.
- **classify_failure (bbf5325):** the `"Error: file_edit on "` prefix matches the Display text exactly, all `FailureKind` consumers handle the new variant, and the round-trip test drives the real `record_tool_results` producer — a genuine pin, not a tautology.
- **Doc fix:** "total occurrences anywhere in the path's recent window (not a consecutive run)" — accurate.

Two nits, non-blocking (ledger findings 121, 122):
- `src/tools/file_edit.rs:923` — the test's final assertion re-records evicted p0 while still at capacity, silently evicting p1, the survivor just pinned. Harmless to the outcome; consider asserting the second eviction explicitly.
- `src/agent/mod.rs:1814` — the EditLoop round-trip case sits between Precondition and InvalidArguments; the same diff keeps the enum, Display arms, and classifier arms alphabetical.

src/tools/file_edit.rs
kitaebot: Confirmed — 919ffdd's spec text matches the code, and the benignness argument is sound: the guard records only on failure paths and clears before a successful write, so it never pre-blocks, and a spurious `EditLoop` costs exactly one re-read, which the error text itself instructs. Documentation over serialization is the right call for a heuristic guard. Closing the loop on this finding as fixed.

src/tools/file_edit.rs
kitaebot: Confirmed — the VecDeque order tracking is correct across `record_futile` (push only on new-key insert, under the mutex) and `clear` (retain-out only when the map held the path), and the test pins the survivor's count. One nit on the test itself in the new review: the final `record_futile("p0", 7)` re-tracks the evicted path while still at capacity, silently evicting p1 — the survivor just pinned. Worth asserting the second eviction explicitly. Finding closed as fixed.

src/error.rs
kitaebot: Confirmed — the `"Error: file_edit on "` prefix matches the Display text exactly, all `FailureKind` consumers handle the new variant, and the round-trip test drives the real `record_tool_results` producer rather than asserting on a hand-built string, so it's a genuine pin. Finding closed as fixed.

src/tools/file_edit.rs
kitaebot: Confirmed — "total occurrences anywhere in the path's recent window (not a consecutive run)" is accurate and consistent with the `EditFutility` doc. Finding closed as fixed.

kitaebot (approved):
Verdict: **correct** — both nits addressed, no new problems.

- The eviction test now asserts the cascading second eviction explicitly: after re-tracking p0 at capacity, `record_futile("p1", 7)` returning 1 (not 3) proves p1 was evicted as the next-oldest by first-record order. The count is right for a fresh single-occurrence ring, and the new comment names the FIFO semantics accurately.
- The round-trip case reordering (Blocked, EditLoop, Precondition, InvalidArguments, …) matches the `ToolError` enum declaration order and the `classify_failure` arm order — the faithful realization of the alphabetical-consistency ask given no CommandFailed case exists in the list.

Delta touches only the two test sites; no production changes. Both prior nits closed as fixed.

Issue #109, opened 2026-08-27
Price cache savings from OpenRouter's live endpoint rates instead of hand-config


amackillop:
## Problem

The cache-savings estimate (#104) prices cached tokens from hand-maintained `[usage.rates]` config. The real rates are knowable: `GET openrouter.ai/api/v1/models/{model}/endpoints` is unauthenticated, on the same host as all provider traffic (already egress-allowlisted), and returns per-endpoint `prompt` / `input_cache_read` / `completion` pricing. The spread across endpoints is wide (GLM 5.2 today: $0.50–$0.68/M input, fp8 vs fp4 quantization), and which endpoint applies is decided by sticky session routing, so a single configured number is a guess. Every chat response already names the endpoint that served it in a `provider` field we currently ignore.

## Proposal

Two commits:

1. **Record the serving endpoint per turn.** Parse the response's `provider` field into `CallUsage`, carry the turn's last-seen value through `TurnMeter`, persist as a nullable `provider` column (migration 0005, same era-boundary pattern as 0002–0004). Independently useful: it surfaces in `/usage` whether traffic is landing on a degraded-quantization endpoint.

2. **Price rows from the live endpoint list.** `/usage` fetches the model's endpoint pricing (short timeout, one fetch per invocation), prices each row by its recorded `(model, provider)`; configured `[usage.rates]` overrides win; fetch failure or an unrecorded/unknown endpoint falls back to config, then to the existing `(N turns unpriced)` count. The report never breaks offline.

Today's prices applied to historical rows is an accepted imprecision — the line is an estimate and says so; recording the provider at turn time is what keeps it grounded.

Issue #101, opened 2026-08-26
Speculative compaction: arm the summary in the background before the threshold


amackillop:
## Problem

Compaction is synchronous at turn boundaries (spec 14): the turn that crosses the threshold stalls for a full summarizer round-trip before work continues. On long tasks this lands repeatedly, always at the worst moment — mid-flow with a full context.

## Proposal

When context enters a band just below the threshold (e.g. within `max(8k, 12.5% of threshold)` tokens), start summarization in the background against a snapshot of the intended prefix. When the threshold is actually crossed, commit the armed result instantly — the summarizer latency is hidden inside turns that were running anyway. Invalidate the armed summary if the intended cut point no longer matches the live history.

## Design forks (why needs-plan)

1. **Snapshot identity**: how to cheaply detect that the armed summary's cut point still matches (prefix hash? entry ids?).
2. **Concurrency**: LCM sessions are single-writer; the background summarize must read a stable snapshot while the turn appends.
3. **Waste bound**: cap on speculative summaries per session so a hovering context size doesn't re-arm forever.
4. Composes with the cache-prefix-aligned summarizer request — speculative or not, the request shape should be the same.

Issue #100, opened 2026-08-26
Compaction summaries re-upload the session at fresh-input rates; ride the main session's cache prefix


amackillop:
## Problem

The summarizer builds a fresh prompt from session content and routes it under its own fixed `summarizer` session_id. Every compaction therefore bills the summarized span at the uncached input rate — content the main session already has sitting in the provider cache at ~5× cheaper read pricing (GLM implicit prefix caching, no write premium).

## Proposal

Build the summarization request as the main session's exact message prefix plus one trailing user instruction ("produce a compaction summary of ..."), routed with the **main session's** session_id. Prefix matching then bills the history as cache reads; only the trailing instruction is fresh input. The response commits as the compaction entry exactly as today; the main session's prefix is untouched, so its own cache survives.

Constraint: the request must be byte-identical to the live prefix up to the trailing message — any assembly divergence (different system prompt, tool schema ordering) silently forfeits the whole discount. Add a debug assertion comparing against the last prompt actually sent.

Scope: live-session compaction only. Distillation over idle sessions gains nothing — their cache entries have expired by the time the duty runs.

Issue #99, opened 2026-08-26
LCM: supersede stale reads and elide zero-signal results, gated on cache economics


amackillop:
## Problem

Long sessions accumulate dead weight that the agent loop re-bills on every iteration: old reads of files that were later re-read or modified (stale content is also a correctness hazard — the model can anchor an edit on a version that no longer exists), and zero-signal tool results (empty grep/glob, no-match searches). The LCM externalizes large outputs at emission (>4096 tokens) but never revisits history afterwards. Recent builds average ~2M prompt tokens per turn across loop iterations; even at cache-read rates, prompt volume is the dominant cost line.

## Proposal

Three additive rules in the LCM:

- **Supersede reads.** When a file is re-read or modified, blank earlier reads of it down to a one-line placeholder. Unconditional — this is correctness first, cost second.
- **Zero-signal elision.** `ToolResult` gains a self-declared zero-signal flag (empty search, no matches). Flagged results collapse to one line. Errors are never flagged — a failure is signal.
- **Cache-aware timing gate.** GLM caching is implicit prefix-matching: cache reads bill ~5× cheaper than fresh input, storage/writes are free. Any in-place history mutation invalidates the cached suffix, which then re-bills at the fresh rate. So mutate only when the suffix after the mutation point is small (≤ ~8k tokens) or the session has idled past the cache TTL (empirically minutes < TTL < 2h). Floor: never blank a result under ~50 tokens — the placeholder wouldn't pay for itself.

Candidate selection is a pure function over history + token estimates; test it directly, verify effect via observed prompt_tokens (spec 14).

PR #111, opened 2026-08-27
Record and report prompt-cache hits in /usage

### LINKED ISSUE #104: /usage: split cached vs fresh input and report cache savings
## Problem

`/usage` reports cost, turns, and tokens, but nothing about cache behavior — so none of the planned cache-economy changes (history elision timing, prefix-aligned summaries, speculative compaction) can be validated after shipping. The provider already parses `prompt_tokens_details.cached_tokens` (spec 02); it's logged at debug level and dropped.

## Proposal

Thread `cached_tokens` through TurnMeter into the usage db: per-task and per-build columns for fresh input, cached input, and output tokens. Derive:

- **cache hit rate** = cached / (cached + fresh input)
- **cache savings** = cached × (input_rate − cache_read_rate), from configured per-model rates. Allow it to go negative so a future policy that pays cache-write premiums reports honestly instead of flattering itself.

Rates vary by pinned OpenRouter provider (GLM 5.2 spread is wide across the 30 endpoints), so rates belong in config next to the model override, not hardcoded.

This is the measurement prerequisite for the other cache issues — land it first, capture a baseline week, then judge each change against it.

amackillop:
Closes #104.

The provider parses `prompt_tokens_details.cached_tokens` from OpenRouter but logged it at debug level and dropped it, so nothing could say how much of a turn's prompt volume the cache absorbed. That number is the baseline the planned cache-economy work (#99, #100, #101) gets judged against.

Three commits, one layer each:

1. **Persist prompt-cache hits per turn** — carry the value through `CallUsage` and `TurnUsage` into a nullable `cached_tokens` column (migration 0004). `Option` end to end: a response without details means the provider did not say, distinct from a reported cold prompt of zero; NULL in the ledger, same contract as `cost`.
2. **Report the cache hit rate in /usage** — `Cache` column on By Build and By Model plus the overall rate in the header. The denominator is the prompt tokens of only the rows that recorded cache details, so pre-migration history cannot dilute the rate toward zero; unmeasured groups render `-`.
3. **Estimate cache savings from configured rates** — new `[usage.rates]` config (USD per 1M tokens per model id) prices cached tokens at the spread between input and cache-read rates. Signed, so a losing policy reports negative. Models missing from the map count as `(N turns unpriced)` rather than being skipped. Empty map (the default) omits the line.

Rates stay operator-supplied for now because the correct value depends on which OpenRouter endpoint sticky routing pinned; #109 covers recording the serving endpoint per turn and pricing from the live endpoint list.

Verify after deploy: any turn, then `/usage` — `Cache` column and header rate appear immediately; the savings line appears once `settings.usage.rates` carries real endpoint numbers.

### DIFF
--- a/src/state_db/migrations/0004_cached_tokens.sql
+++ b/src/state_db/migrations/0004_cached_tokens.sql
@@ -1,0 +1,4 @@
+-- Prompt-cache hits per turn (spec 27). Nullable three ways: rows
+-- predating the column, and turns whose provider never reported
+-- prompt_tokens_details. NULL renders as "-", never zero.
+ALTER TABLE turns ADD COLUMN cached_tokens INTEGER;
--- a/src/usage.rs
+++ b/src/usage.rs
@@ -12,9 +12,12 @@ use std::cmp::Ordering;
 use std::fmt::{self, Write as _};
 use std::sync::{Arc, Mutex};
 
+use std::collections::HashMap;
+
 use rusqlite::{Connection, params};
 
 use crate::agent::envelope::ChannelSource;
+use crate::config::ModelRates;
 use crate::state_db::StateDb;
 use tracing::warn;
 
@@ -91,13 +94,21 @@ pub struct TurnRecord<'a> {
 /// migration.
 pub struct UsageLedger {
     conn: Arc<Mutex<Connection>>,
+    /// Operator-supplied prices for the report's savings estimate;
+    /// carried here so command sites need no extra config plumbing.
+    rates: HashMap<String, ModelRates>,
 }
 
 impl UsageLedger {
-    pub fn new(db: &StateDb) -> Self {
+    pub fn new(db: &StateDb, rates: HashMap<String, ModelRates>) -> Self {
         Self {
             conn: db.connection(),
+            rates,
         }
+    }
+
+    pub fn rates(&self) -> &HashMap<String, ModelRates> {
+        &self.rates
     }
 
     /// Read every recorded turn, projected to the columns the report
@@ -211,6 +222,7 @@ impl Agg {
     /// Share of prompt tokens served from cache, over the rows that
     /// reported cache details; `None` when none did.
     fn cache_rate(&self) -> Option<f64> {
+        // Token sums sit far below f64's 2^53 integer ceiling.
         #[allow(clippy::cast_precision_loss)]
         (self.cached_prompt > 0).then(|| self.cached as f64 / self.cached_prompt as f64)
     }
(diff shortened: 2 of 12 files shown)


src/state_db/migrations/0004_cached_tokens.sql
```diff
@@ -0,0 +1,4 @@
+-- Prompt-cache hits per turn (spec 27). Nullable three ways: rows
```
kitaebot, line 1: The comment says "Nullable three ways:" but lists only two origins (rows predating the column; providers that never reported `prompt_tokens_details`). NULL is a single state regardless of origin — either name the intended third origin or drop the count.

src/usage.rs
```diff
@@ -194,9 +213,26 @@ impl Agg {
...
+    fn cache_rate(&self) -> Option<f64> {
+        // Token sums sit far below f64's 2^53 integer ceiling.
+        #[allow(clippy::cast_precision_loss)]
+        (self.cached_prompt > 0).then(|| self.cached as f64 / self.cached_prompt as f64)
     }
 }
 
+/// `87%`, or `-` when the group has no measurable rows.
+fn fmt_rate(rate: Option<f64>) -> String {
+    rate.map_or_else(|| "-".to_string(), |r| format!("{:.0}%", r * 100.0))
```
kitaebot: `{:.0}%` rounding can render a sub-100% hit rate as "100%" (e.g. 1499/1500 cached), which reads like a guarantee of full cache coverage. Display-only; a floor or one decimal would avoid the boundary lie.

kitaebot (approved):
Verdict: **correct**. All three layers implement what their commit messages claim.

The `Option`/NULL contract is preserved end to end — provider `prompt_tokens_details` maps to `cached_tokens: Option<u32>` only when details were present, `TurnUsage::add_call` keeps `None` until a call reports (pinned by the mixed-`None` test), SQL stores NULL, `rows()` restores `Option<u64>`, and the report renders absent data as `-`, never zero. Migration 0004 is an additive nullable column, appended last, safe on populated ledgers. The rate arithmetic checks out: the denominator is prompt tokens of cache-reporting rows only (test pins 800/1000, not 800/1500, so legacy history cannot dilute the rate), savings are signed, unpriced models surface as `(N turns unpriced)` instead of vanishing, and an empty rates map omits the line. Spec 27's updated text matches the implementation, and the README documents the new config with honest commentary on rate variability.

Two nits, neither blocking (ledger findings 119, 120):
- `0004_cached_tokens.sql:1` — the comment says "Nullable three ways" but lists two origins; NULL is one state regardless of origin.
- `src/usage.rs:233` — `{:.0}%` rounding can render a sub-100% hit rate as "100%" (e.g. 1499/1500). Display-only boundary lie; a floor or one decimal would avoid it.

Findings stay undispositioned pending the author's response.

amackillop:
Findings addressed:

- **120** (rate rounding): fixed in 16c495d — the displayed percentage is floored, so 100% now appears only when every measured prompt token came from cache. Boundary pinned by a test (1499/1500 renders 99%).
- **119** (miscounted "three ways" in the 0004 comment): acknowledged, not fixed. The pre-commit hook enforces migrations as append-only with no released/unreleased distinction, and a comment typo does not justify bypassing that fence. If it grates, the original commit can be rebased after review concludes; the hook accepts the file as newly added in a rewritten commit.

kitaebot (approved):
Verdict: **correct** — finding 120 fully resolved, no new problems.

The floor is correct on every reachable input: the only producers are `Agg::cache_rate()` calls guarded by `cached_prompt > 0` with u64 sums far below 2^53, so the rate is always finite and non-negative — 0.0 renders "0%" (measured-zero, distinct from the unmeasured "-"), 1499/1500 renders "99%", and "100%" now appears iff every measured prompt token came from cache. The new test pins exactly the discriminating case the old rounding got wrong. No other behavior changed: floor is identity on integral percentages.

On 119 (the "three ways" comment typo): the author's refusal is justified. The pre-commit hook rejects any modification to applied migration files with no released/unreleased distinction, so the fix would mean `--no-verify` on the very fence that protects applied user databases, or a mid-review history rewrite. Acknowledged no-action with a legitimate post-review path (rewrite the original commit, where the hook sees the file as newly added). Fine for a comment-only nit.

PR #115, opened 2026-08-28
Price cache savings from OpenRouter's live endpoint rates

### LINKED ISSUE #109: Price cache savings from OpenRouter's live endpoint rates instead of hand-config

amackillop:
Closes #109.

The cache-savings line (#111) priced cached tokens from hand-maintained `[usage.rates]` config, but a single per-model number is a guess: OpenRouter's per-endpoint spread for GLM 5.2 alone runs $0.50 to $0.68 per million input tokens, and which endpoint applies is decided by sticky session routing. Both halves of the real answer are available: every chat response names its serving endpoint in a `provider` field, and `GET /api/v1/models/{model}/endpoints` lists each endpoint's prices, unauthenticated, on a host already egress-allowlisted.

Two commits:

1. **Record the serving endpoint per turn** — parse the response's `provider` field through `CallUsage`/`TurnUsage` (last-seen-wins: a mid-turn endpoint change is failover, and a provider-less call never erases a known name) into a nullable `provider` column (migration 0005). Independently useful: it surfaces whether traffic is landing on a degraded-quantization endpoint. `CallUsage`/`TurnUsage` lose `Copy` for the `String`; all use sites already moved by value.
2. **Price rows from the live endpoint list** — `/usage` fetches endpoint pricing for each model that needs it (has cache data, has a recorded endpoint, no config entry) and prices each row by its `(model, provider)` pair. Resolution order: config override → live endpoint rate → `(N turns unpriced)`. Fetch failure degrades to config-then-unpriced with a warning, never a failed report; `Api::Custom` returns no pricing endpoint, so e2e fixture runs never fetch. An endpoint listing no cache-read price bills hits at the full input rate — kept as a real spread of zero, not dropped.

Accepted imprecision, stated in spec 27: live rates are today's prices applied to historical rows — same limitation config had; the recorded endpoint is what keeps the estimate grounded.

After deploy: nothing to configure. The savings line prices itself for turns recorded after migration 0005; `[usage.rates]` becomes a pure override knob.

### DIFF
--- a/src/clients/openrouter_pricing.rs
+++ b/src/clients/openrouter_pricing.rs
@@ -92,11 +92,24 @@ impl PricingClient {
 }
 
 /// Parse a raw endpoints response into per-endpoint rates. Pure.
+/// Status routing mirrors `chat_completion::interpret_response`: the
+/// error variant is the failure mode, not a catch-all label.
 fn interpret_pricing(raw: &RawResponse) -> Result<Vec<EndpointRates>, ProviderError> {
     debug!(status = raw.status, "Endpoint pricing response");
-    if !(200..=299).contains(&raw.status) {
-        let body = String::from_utf8_lossy(&raw.body);
-        return Err(ProviderError::BadRequest(format!("{}: {body}", raw.status)));
+    match raw.status {
+        200..=299 => {}
+        429 => return Err(ProviderError::RateLimited),
+        500..=599 => {
+            let body = String::from_utf8_lossy(&raw.body);
+            return Err(ProviderError::ServerError(format!(
+                "{}: {body}",
+                raw.status
+            )));
+        }
+        s => {
+            let body = String::from_utf8_lossy(&raw.body);
+            return Err(ProviderError::BadRequest(format!("{s}: {body}")));
+        }
     }
     let parsed: PricingResponse = serde_json::from_slice(&raw.body)
         .map_err(|e| ProviderError::InvalidResponse(e.to_string()))?;
@@ -198,6 +211,24 @@ mod tests {
     fn non_2xx_is_an_error_carrying_the_body() {
         let err = interpret_pricing(&raw(404, "no such model")).unwrap_err();
         assert!(err.to_string().contains("404"), "{err}");
+    }
+
+    /// The variant names the failure mode: a rate limit or server
+    /// error must not be labeled a bad request.
+    #[test]
+    fn status_routes_to_the_matching_variant() {
+        assert!(matches!(
+            interpret_pricing(&raw(429, "")).unwrap_err(),
+            ProviderError::RateLimited
+        ));
+        assert!(matches!(
+            interpret_pricing(&raw(500, "boom")).unwrap_err(),
+            ProviderError::ServerError(s) if s.contains("boom")
+        ));
+        assert!(matches!(
+            interpret_pricing(&raw(404, "")).unwrap_err(),
+            ProviderError::BadRequest(_)
+        ));
     }
 
     #[tokio::test]
--- a/src/clients/openrouter_pricing.rs
+++ b/src/clients/openrouter_pricing.rs
@@ -1,0 +1,212 @@
+//! `OpenRouter` endpoint-pricing client (spec 27).
+//!
+//! `GET {base}/{model}/endpoints` lists every upstream endpoint
+//! serving a model with its prices. Unauthenticated, same host as the
+//! chat API, so it rides the existing egress allowlist. Pure parsing
+//! lives in [`interpret_pricing`]; IO is a stored closure, mirroring
+//! [`super::chat_completion::CompletionsClient`].
+
+use std::future::Future;
+use std::pin::Pin;
+use std::sync::Arc;
+use std::time::Duration;
+
+use serde::Deserialize;
+use tracing::{debug, error};
+
+use super::RawResponse;
+use crate::config::ModelRates;
+use crate::error::ProviderError;
+
+type GetResult = Result<RawResponse, ProviderError>;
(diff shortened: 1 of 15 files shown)


src/clients/openrouter_pricing.rs
```diff
@@ -0,0 +1,212 @@
...
+    pub async fn endpoint_rates(&self, model: &str) -> Result<Vec<EndpointRates>, ProviderError> {
+        let raw = (self.get)(format!("{model}/endpoints")).await?;
+        interpret_pricing(&raw)
+    }
+}
+
+/// Parse a raw endpoints response into per-endpoint rates. Pure.
+fn interpret_pricing(raw: &RawResponse) -> Result<Vec<EndpointRates>, ProviderError> {
+    debug!(status = raw.status, "Endpoint pricing response");
+    if !(200..=299).contains(&raw.status) {
```
kitaebot: All non-2xx statuses map to `ProviderError::BadRequest`, whose docstring defines it as HTTP 4xx other than 401/403/429. A 429 or 5xx from the pricing endpoint logs as "Provider rejected the request: 500: …" — a label that lies about the failure mode on an already-degraded path. Mirror `chat_completion::interpret_response`'s 429 → `RateLimited` / 5xx → `ServerError` split so the warn line names the real failure.

kitaebot (approved):
Verdict: **correct** — both commits do what their messages claim.

Provider plumbing: `ChatResponse.provider` is `#[serde(default)]` so non-OpenRouter APIs parse; the field lifts into `CallUsage` even without a usage block; `TurnUsage::add_call` implements last-seen-wins with `None` never erasing a known name (tested both directions); migration 0005 is a single additive nullable column consistent with #111's NULL-means-absence contract. Live pricing: `Api::pricing_endpoint()` returns `Some` only for OpenRouter so `Custom` fixtures never fetch; the client sends a bare GET (unauthenticated claim true in code) with a 5s timeout; fetch failure degrades to a warn plus partial map, never a failed report; resolution is config → live → unpriced per `(model, provider)`; zero-spread endpoints are kept and distinguishable from unpriced; `report()` stays pure. Spec 27 and the README match the implementation.

Process note: although the dispatch framed this as a re-review, no prior review of this PR was ever published — my first reviewer dispatch was interrupted before a verdict came back. This is the initial review of the full PR.

One nit, non-blocking (ledger finding 128): `interpret_pricing` maps every non-2xx status — including 429 and 5xx — to `ProviderError::BadRequest`, whose own docstring defines it as HTTP 4xx. A rate-limited or server-error response logs as "Provider rejected the request: 500: …", a label that lies about the failure mode. The sibling `chat_completion::interpret_response` splits 429 → `RateLimited` and 5xx → `ServerError`; this client should do the same.

kitaebot (approved):
Re-review of the rebase delta (a77fba9 → 38a00ed): **correct**.

The PR was rebased onto newer master; its own two commits are unchanged in content. The delta is purely mechanical adaptation to master's 1419fe7: `retry_after_secs: None` initializations at the PR's client and test construction sites, with an accurate comment that the pricing client never retries. Verified no interaction problems: `RawResponse.retry_after_secs` is consumed only by the GitHub client (the scope of 1419fe7); the pricing path reads only status/body, makes one attempt per model, and degrades to a warn — no retry loop exists for the `None` to affect. Migration 0005 stays append-only-consistent (no inherited commit added a migration). The provider recording also composes correctly with the chat retry loop: the provider is taken from the successful attempt, so a 429-then-success sequence records the endpoint that actually served the turn.

The prior nit (finding 128, `interpret_pricing` mapping 429/5xx to `BadRequest`) is still present, unchanged, and still non-blocking — left to the author.

amackillop:
Finding 128 addressed in d7ea1d1: interpret_pricing now routes status by failure mode, mirroring chat_completion::interpret_response — 429 to RateLimited, 5xx to ServerError with the body preserved, other 4xx to BadRequest. All three variants pinned by a test.

src/clients/openrouter_pricing.rs
kitaebot: Confirmed fixed in d7ea1d1 — verified the routing matches `chat_completion::interpret_response` arm-for-arm, with all three variants pinned by `status_routes_to_the_matching_variant` (including body preservation on `ServerError`). Closing this out.

PR #117, opened 2026-08-28
Ride the main session's cache prefix for compaction summaries

### LINKED ISSUE #100: Compaction summaries re-upload the session at fresh-input rates; ride the main session's cache prefix

amackillop:
Closes #100.

Leaf-pass compaction re-uploaded every summarized span at fresh-input rates under the fixed `summarizer` session key, even though the identical bytes sit in the provider's implicit prefix cache under the main session at ~5x cheaper read pricing.

Each leaf chunk is now summarized as: the last request the agent loop actually sent, truncated right after the chunk, plus one trailing user instruction scoping "the final N messages" — routed with the main session's sticky key, the snapshot's tool schemas, and the **main** provider. Only the instruction is fresh input. The response commits as the summary row exactly as before; the session's own prefix is untouched.

Per chunk this trades chunk-tokens-at-fresh-rate for prefix-tokens-at-cached-rate, a win while a cycle stays under ~10 chunks; the deployed shape (~5 chunks of 20k) is roughly 44% cheaper on compaction input, and the summary comes from the main model with the full conversation in view instead of the flash override seeing a bare segment.

Commits, in review order:

1. **Generalize the escalation ladder over a call closure** — the L1/L2/L3 ladder no longer hard-wires the fresh-prompt path; the caller decides how the chunk reaches the model.
2. **Observe the request each completion actually sends** — the deviation from the issue text: instead of a debug assertion comparing a re-assembled prompt against the last one sent, the LCM engine keeps a byte-exact `SentRequest` snapshot of what actually went out. Byte identity holds by construction; misalignment falls back at runtime instead of asserting.
3. **Ride the main session's cache prefix in the leaf pass** — guards, in order, each falling back to the fresh path: snapshot exists → same session → younger than 2 minutes (cold rides re-upload the whole prefix at fresh rates, dearer than the path being replaced) → chunk aligns byte-exactly in the snapshot. Alignment self-scopes riding to the leaf pass: condensed chunks are synthetic summaries absent from any sent request. A ride ladder that bottoms out at level-3 truncation retries the fresh ladder, so riding never yields a worse summary.
4. **Document cache-prefix riding in specs 14 and 02.**

Policy decisions taken during planning: riding wins over `model_overrides.summarizer` for live leaf compaction (the cache lives under the main model; the override keeps serving ingest exploration, the condensed pass, and every fallback), and the age guard is a 2-minute constant, not config.

Live verification after deploy:
- Push a session past the soft threshold and let turn-boundary compaction fire; `vm-logs` shows "summarizing chunk via the main session's cache prefix" per leaf chunk.
- OpenRouter activity shows the compaction calls under the session's id with `cached_tokens` ≈ prefix size.
- A manual `/compact` more than 2 minutes after the last turn logs the fallback and behaves exactly as today.


specs/14-context-engine.md
```diff
@@ -692,7 +701,10 @@ The `model` column on the summary records which LLM produced it. If a
 dedicated summarizer model is configured (`provider.model_overrides.summarizer`),
 it is used instead of the agent's primary model. This allows using a
 cheaper/faster model for summarization. Level 3 records
-`level3-truncate` (no LLM).
+`level3-truncate` (no LLM). Exception: leaf chunks summarized by
+riding the main session's cache prefix (§"Cache-Prefix Riding") go
+through the **main** model deliberately — the cache being ridden lives
+under it, and a summarizer override does not apply there.
```
kitaebot: The `model` column on the summary records which LLM produced it. If a
dedicated summarizer model is configured (`provider.model_overrides.summarizer`),
it is used instead of the agent's primary model. This allows using a
cheaper/faster model for summarization. Level 3 records
`level3-truncate` (no LLM). Ridden leaf chunks go through the **main**
model deliberately — the cache being ridden lives under it, and a
summarizer override does not apply there (§"Cache-Prefix Riding").
The `model` column records the escalation level (`level1`, `level2`,
`level3-truncate`), not the model name; ridden and fresh-path
summaries are indistinguishable in the table.

src/context/lcm/compaction.rs
```diff
@@ -13,16 +13,105 @@ use std::sync::{Arc, Mutex};
...
+/// `sent[..end]` is the shortest sent prefix that ends with the chunk.
+/// `None` when the chunk does not appear — condensed chunks (synthetic
+/// `<summary>` messages built this cycle) always land here, which is
+/// what scopes riding to the leaf pass.
+pub(super) fn prefix_end(sent: &[Message], chunk: &[Message]) -> Option<usize> {
+    if chunk.is_empty() || chunk.len() > sent.len() {
+        return None;
+    }
+    sent.windows(chunk.len())
+        .position(|window| window == chunk)
```
kitaebot: `position()` takes the *first* matching window. When the chunk's message run appears twice in the sent request — a repeated exchange, or an older span whose summary quotes the session verbatim — the ride truncates after the earlier occurrence, so the "final N messages" instruction scopes the wrong span, and the bad summary passes the length thresholds silently. `rposition` (last match) matches the "last request, truncated right after the chunk" intent, whose tail is the most recent history:

```suggestion
    sent.windows(chunk.len())
        .rposition(|window| window == chunk)
        .map(|start| start + chunk.len())
```

src/agent/mod.rs
```diff
@@ -796,6 +799,8 @@ async fn squeeze(
     .await?
     .map_err(Error::Provider)?;
 
+    engine.observe_request(assembled.messages, Arc::from([]));
```
kitaebot, line 802: `Arc::from([])` overwrites the snapshot's real tool schemas with an empty list. A turn ending via squeeze leaves a snapshot whose tools diverge from what the session's cache prefix was built with, so the next compaction rides without schema byte-identity and silently forfeits the discount. Pass the turn's `tool_definitions` into `squeeze` (as the turn loop does at agent/mod.rs:538) and observe with them, or skip observing here since the squeeze request adds nothing the next ride needs.

kitaebot (commented):
The core mechanism is sound: the escalation-ladder generalization is behavior-preserving on the fresh path, `prefix_end` locates the chunk as a contiguous run and truncates right after it, the guard ladder (snapshot present → session match → 2-minute age → per-chunk byte alignment) matches spec 14's ordering, and the ride-ladder-exhaustion retry means the fresh ladder always gets its turn before the lossy floor is accepted.

Three gaps remain — none blocks, but all are worth addressing:

1. **Spec/code mismatch on the `model` column** (specs/14-context-engine.md:700, compaction.rs:268). The paragraph this PR edits still claims the `model` column "records which LLM produced it" and now documents that ridden leaf chunks go through the main model — but `write_leaf_summary` writes `outcome.level.tag()` (`level1`/`level2`/`level3-truncate`), never a model name. After this change a ridden summary and a fresh-path summary are indistinguishable in the `summaries` table, and the sentence is a claim the code was just re-audited against and still doesn't meet. Either record the actual model (or tag ridden outcomes distinctly), or reword the sentence.

2. **First-match window search** (compaction.rs:41). `prefix_end` takes the first matching window of the chunk inside the snapshot. When the same message run appears twice in the sent request — a repeated exchange, or an older span whose summary quotes the session verbatim — the ride truncates after the earlier occurrence, so the "final N messages" instruction scopes the wrong span, accepted silently because the output still passes the length thresholds. `rposition` (last match) matches the "last request, truncated right after the chunk" intent; at minimum this deserves a comment acknowledging the choice.

3. **Squeeze overwrites the snapshot's tool schemas** (agent/mod.rs:802). `squeeze` records its no-tools completion into the snapshot with `Arc::from([])`, clobbering the snapshot that carried the real tool schemas. A turn that ends via the state-report or confirm-reply squeeze leaves a snapshot whose tools no longer match what the main session's cache was built with, so the next turn-boundary compaction rides with an empty schema list — forfeiting the byte-identity the mechanism exists to preserve. The turn loop passes the real `tool_definitions` (agent/mod.rs:538); `squeeze` should receive and pass them too, or skip observing there.

One test-coverage note: the ride path's distinguishing logic — per-chunk alignment fallback, ladder-exhaustion retry to fresh, condensed-pass self-scoping — is only exercised end-to-end through `force_compact`; a `summarize_chunk`-level test with a ride ladder that bottoms out at `Deterministic` would pin the "never worse than the fresh path" guarantee the commit message makes.

specs/14-context-engine.md
amackillop: Fixed in a29fbc0 — reworded to what the code does. The "records which LLM produced it" claim predates this PR and was already false (write_leaf_summary has always stored the level tag); the paragraph now says the column records the escalation level, that the actually-running model is unrecorded, and that ridden and fresh-path summaries are indistinguishable in the table with the ride decision living in the logs.

src/context/lcm/compaction.rs
amackillop: Applied in df8dba4, with the intent comment and a duplicate-run test. One correction to the failure scenario: a colliding window requires byte-equal whole messages, so every occurrence is identical content and the produced summary is the same either way — a summary quoting the session can't collide because summaries render inside <summary> wrappers as single messages. rposition is still the right choice: the last occurrence is the run whose surrounding context the instruction describes.

src/agent/mod.rs
amackillop: Disagree — no change. The snapshot's contract is byte-identity with the request that most recently populated the cache, not with "the session's tool-ful cache". Implicit prefix caching has no single per-session entry: every request's rendered prompt is cached until TTL, and a ride only needs identity with some warm entry. The squeeze request itself just uploaded a no-tools prompt (that fresh upload is pre-existing squeeze behavior, untouched by this PR) and thereby created exactly such an entry; a ride built from its snapshot re-sends those bytes — same messages, tools=[], same session key — and hits it. Self-consistent by construction, discount intact.

Both proposed alternatives are worse or equal: passing the turn's tool_definitions into observe while chat sends &[] would record bytes that were never sent — the exact divergence this mechanism exists to prevent; skipping the observation would ride the older tool-ful entry, which is equally warm and no cheaper. Since both squeeze call sites are turn-terminal, the empty-tools snapshot is also the newest entry, i.e. the one that expires last.

amackillop:
Review addressed: a29fbc0 (spec reword — model column records escalation level, not an LLM), df8dba4 (rposition + duplicate-run test), effb936 (summarize_chunk-level test pinning the ladder-exhaustion fresh retry). Finding 3 (squeeze snapshot tools) contested on the thread — the snapshot is byte-consistent with the squeeze request's own cache entry, so nothing is forfeited.

specs/14-context-engine.md
kitaebot: Confirmed — a29fbc0 reads right. Verified `write_leaf_summary` stores the level tag and that `/stats` counts `model = 'level3-truncate'` rows exactly as the reworded paragraph describes, so the spec now matches the code. Fair point that the stale claim predates the PR; my finding was that this PR edited the paragraph while leaving the false sentence in, and the reword resolves it either way.

src/context/lcm/compaction.rs
kitaebot: Confirmed in df8dba4 — `rposition` with the intent comment, and `prefix_end_takes_the_last_of_duplicate_runs` asserts `Some(4)` where the old `.position()` would have returned `Some(2)`, so the test pins the change.

Your correction of my failure scenario is right, and I retract it: window equality compares whole `Message` values, so a colliding occurrence is byte-identical content and the summary text is unaffected either way — only the surrounding context the trailing instruction describes differs, which is exactly why the last occurrence is the correct anchor. The "older summary quoting the session verbatim" case can't collide at all, since summaries render inside `<summary>` wrappers as single messages.

src/agent/mod.rs
kitaebot: You've convinced me — concession on the finding, with the verification that got me there:

The snapshot contract is byte-identity with the last request actually sent (`observe_request` records exactly what went to `provider.chat`), and I was wrong to frame the tool-ful turn request as "the" cache entry. Squeeze sends a real no-tools completion under the session key, which populates the cache with exactly those bytes; the ride re-sends them and hits that entry. Self-consistent, discount intact.

I also checked the turn-terminal claim: squeeze's only callers are `final_answer` and `state_report` (via `BudgetPolicy::Fail`), both of which end the turn — so the empty-tools snapshot is the newest entry and the last to expire.

And agreed on both alternatives being worse: observing `tool_definitions` while chat sends `&[]` records bytes that were never sent — the exact divergence the mechanism exists to prevent — and skipping the observation just rides an older, equally warm entry. Withdrawn.

kitaebot (approved):
Re-review of the delta (09f96bb → effb936): verdict correct, all findings addressed.

- a29fbc0 — confirmed fixed: the spec now states the `model` column records the escalation level, matching `write_leaf_summary` and the `/stats` level-3 count.
- df8dba4 — confirmed fixed: `rposition` with an intent comment, pinned by `prefix_end_takes_the_last_of_duplicate_runs` (asserts `Some(4)`; the old `position()` returns `Some(2)`). The author's correction of the original failure scenario is right and retracted.
- effb936 — confirmed fixed: the `summarize_chunk`-level test drives a ride ladder that fails both LLM levels and asserts the fresh ladder's summary wins at exactly two ride calls; without the retry the outcome would be deterministic truncation, so it's not vacuous.
- #131 (squeeze snapshot tools) — withdrawn. The dispute is sound: the snapshot byte-matches the request squeeze actually sent, which itself populates the cache entry the ride then hits; both squeeze call sites are turn-terminal.

PR #118, opened 2026-08-28
Trust devshell-signed exports over stderr noise

### LINKED ISSUE #114: has_nix_error misclassifies devshells whose hook prints non-nix error lines; re-evaluates per git call
## Symptom (2026-08-27 duty journal, CumuloGlobal/lightning-node)

8 WARNs in 37 seconds (17:02:51–17:03:28), all `direnv failed, running git without devshell` from `kitaebot::tools::git::git_cli`, plus one `devshell warm failed`. Each carries the same stderr, whose tail is the tell:

```
lsps1-server/lsps1-http postinstall: Error: EACCES: permission denied, scandir '/lib64'
error: recipe `pnpm-install` failed on line 19 with exit code 1
...
Development environment ready
...
direnv: export +AR +AS +CC ... ~PATH
```

Direnv exited 0 and exported the full environment — but the error text kitaebot attached is `direnv export json reported an error:`, i.e. the `has_nix_error` classification path fired on a successful evaluation.

## Mechanism

- The only `error:`-prefixed line in that stderr is **just's** error format (`error: recipe 'pnpm-install' failed...`), not nix's. The prisma postinstall EACCES on `/lib64` is the known #72 Landlock residual; the repo's shellHook tolerates it and the devshell still comes up ("Development environment ready", full export).
- `src/tools/direnv.rs:289-291` — `has_nix_error` matches ANY line that (after ANSI strip) starts with `error:`. The doc comment says "Nix's error marker is `error:` at the start of a line" — but just, pnpm, cargo, and most build tools also print `error:`-prefixed lines to stderr during a shellHook. The heuristic cannot distinguish nix evaluation failure from a non-fatal hook step failure.
- `src/tools/direnv.rs:366-376` — on exit 0 with a matching line, the working export is discarded and `DirenvError::Failed` returned.
- `src/tools/direnv.rs:271-275` — fast failures are deliberately not cached (`cache.remove(dir)`), so the next caller retries. Combined with per-git-command direnv resolution in git_cli, one repo whose shellHook prints any `error:` line re-runs its **entire devshell evaluation on every git invocation** — 8 full `pnpm -r install` + prisma generate passes in 37 seconds above — and every git command runs without the devshell it was entitled to.

## Cost

- Wasted work: a full devshell build (network fetch + postinstalls) per git command, unbounded, for as long as the repo's hook prints an `error:` line.
- Wrong env: git (and anything else resolved through the cache) silently runs outside the devshell.

## Fix directions (fork needing a call)

1. **Tighten the marker**: nix errors have a distinctive shape (e.g. `error:` followed by a nix error code path, or the flake-attribute form). Fragile against nix output changes.
2. **Trust the export**: if `direnv export json` exited 0 AND produced a non-empty export, treat the evaluation as good regardless of stderr noise; the original bug this heuristic guards (direnv swallowing `use flake` failure and exporting a *bare* environment) is detectable as an export lacking the devshell's signature vars, or simply by exit status + empty export.
3. **Cache failures with a TTL** (like timeouts, `TIMEOUT_TTL`): bounds the re-evaluation storm even if classification stays conservative.

2 + 3 together look right: correctness from trusting the export, waste-bounding from a failure TTL. Not fixed here per triage policy.

Related but separate: the underlying `/lib64` EACCES in the daemon's Landlock ruleset (postinstalls under the daemon grant) is the #72 residual and makes every evaluation of this repo's devshell noisy even after the classification fix.

kitaebot:
Closes #114.

The stderr `error:` check in `evaluate_direnv` ran before the export was parsed, so any shellHook step printing an `error:`-prefixed line (just, pnpm, cargo all do) failed a fully-working devshell. Fast failures are uncached by design, so every git call re-ran the entire devshell evaluation: eight full pnpm-install passes in 37 seconds observed on CumuloGlobal/lightning-node, whose hook tolerates a failing prisma postinstall and still reaches "Development environment ready" with a complete export.

Classification now keys on the authoritative artifact — the export: a nix devshell signature (`IN_NIX_SHELL` or `NIX_STORE`) is present iff the flake actually evaluated, so a signature-bearing export succeeds regardless of stderr noise. Signature-less exports keep the #41 stderr fallback unchanged (bare export + nix `error:` line → `Failed`), which also covers an `.envrc` that sets variables before a failing `use flake`. Verified against the real kitaebot flake: its successful export carries 74 keys including both signature vars, with clean stderr.

Implements the approved plan on #114 (comment 5447407609). The issue's option 3 (`Failed` TTL) was deliberately deferred per plan decision 3. Spec 03 requirement 4 reworded to the signature-first contract; failure-caching semantics (requirement 3) untouched.

Validation: all 25 `tools::direnv` tests pass, including two new ones — the incident shape (signature + just error line → `Ready`) and the regression shape (manual export, no signature, nix error → `Failed`). `nix flake check` passed at commit time (cargo-deny duplicate-crate warnings are pre-existing). The 32 `tools::git::fixup`/`rebase` failures in the in-sandbox suite are pre-existing on master (verified with the change stashed); the nix gate runs them green.

### DIFF
--- a/src/tools/direnv.rs
+++ b/src/tools/direnv.rs
@@ -382,7 +382,10 @@ async fn evaluate_direnv(
 
     // A devshell signature var proves the flake evaluated: nix always
     // exports IN_NIX_SHELL and NIX_STORE from a live devshell, and a
-    // failed evaluation has no devshell environment to export.
+    // failed evaluation has no devshell environment to export. The
+    // subprocess env is env_clear()ed to SAFE_ENV_VARS, which contains
+    // neither var, so the export diff can never omit them as already
+    // present in the parent.
     if !has_devshell_signature(&env) {
         // direnv swallows `use flake` failures, exporting a bare
         // environment; nix's `error:` marker on stderr is the tell.
--- a/src/tools/direnv.rs
+++ b/src/tools/direnv.rs
@@ -363,18 +363,6 @@ async fn evaluate_direnv(
         )));
     }
 
-    // direnv swallows `use flake` failures: it exits 0, prints the nix
-    // error to stderr, and exports a bare environment. Detect this by
-    // checking for nix's `error:` marker on stderr (after stripping ANSI
-    // color codes). On success stderr carries only direnv log messages
-    // (prefixed `direnv:`), which never start with `error:`.
-    let stderr = output.stderr.trim();
-    if has_nix_error(stderr) {
-        return Err(DirenvError::Failed(format!(
-            "direnv export json reported an error: {stderr}"
-        )));
-    }
-
     // direnv outputs nothing when there's no .envrc or it's not allowed.
     let stdout = output.stdout.trim();
     if stdout.is_empty() {
@@ -387,13 +375,32 @@ async fn evaluate_direnv(
 
     // Filter out nulls (variables direnv wants to unset — irrelevant since
     // we build the subprocess env from scratch, not from the current process).
-    let env = raw
+    let env: HashMap<String, String> = raw
         .into_iter()
         .filter_map(|(k, v)| v.map(|v| (k, v)))
         .collect();
 
+    // A devshell signature var proves the flake evaluated: nix always
+    // exports IN_NIX_SHELL and NIX_STORE from a live devshell, and a
+    // failed evaluation has no devshell environment to export.
+    if !has_devshell_signature(&env) {
+        // direnv swallows `use flake` failures, exporting a bare
+        // environment; nix's `error:` marker on stderr is the tell.
+        let stderr = output.stderr.trim();
+        if has_nix_error(stderr) {
+            return Err(DirenvError::Failed(format!(
+                "direnv export json reported an error: {stderr}"
+            )));
+        }
+    }
+
     debug!(dir = %dir.display(), "Direnv evaluation complete");
     Ok(env)
+}
+
+/// Whether the export carries proof that a nix devshell evaluated.
+fn has_devshell_signature(env: &HashMap<String, String>) -> bool {
+    env.contains_key("IN_NIX_SHELL") || env.contains_key("NIX_STORE")
 }
 
 #[cfg(test)]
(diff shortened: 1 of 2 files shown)


src/tools/direnv.rs
```diff
@@ -387,15 +375,34 @@ async fn evaluate_direnv(
 
     // Filter out nulls (variables direnv wants to unset — irrelevant since
     // we build the subprocess env from scratch, not from the current process).
-    let env = raw
+    let env: HashMap<String, String> = raw
         .into_iter()
         .filter_map(|(k, v)| v.map(|v| (k, v)))
         .collect();
 
+    // A devshell signature var proves the flake evaluated: nix always
```
amackillop, line 383: **should-fix:** this comment states an invariant the tooling does not deliver, and the plan's assumption inherits the error. nix does set these vars in every devshell — but `direnv export json` emits a *diff* against the calling process's environment, and a var whose value already matches the parent is omitted. `IN_NIX_SHELL=impure` and `NIX_STORE=/nix/store` are the same everywhere, so run the daemon from inside any nix shell and a **successful** eval exports no signature, falls to the stderr path, and the #114 false-fail returns — exactly when someone tests this fix from `nix develop`. Production (systemd service, no nix parent) is unaffected and the degradation is to pre-PR behavior, hence no block.

Fix the comment and spec 03's "proves the flake evaluated" to state the real constraint — it holds while the daemon itself runs outside any nix shell — or discriminate on a signal that survives the diff, e.g. `PATH` gaining a `/nix/store` segment.

amackillop (commented):
Verdict: **correct** — the mechanism is sound, the #41 fence survives via the signature-less fallback, and both incident shapes are pinned by tests. One should-fix inline. The manual-export misclassification is consciously accepted in the plan on #114, so it is not raised here.

One observation, no action: reordering puts the stderr check after the empty-stdout branch, so "exit 0 + empty stdout + nix error" now reads as no-devshell instead of Failed. Traced it — the swallow case always exports at least direnv's own vars, so the combination looks unreachable. Noting it so nobody rediscovers it as a mystery.

src/tools/direnv.rs
kitaebot: The scenario doesn't reach the subprocess. The direnv subprocess doesn't inherit the daemon's environment: `cli_runner` runs it with `env_clear()` and then `.envs(safe_env())`, and `SAFE_ENV_VARS` (src/tools/mod.rs:58) contains neither `IN_NIX_SHELL` nor `NIX_STORE`. So even with the daemon started from inside `nix develop`, the export diff is computed against a scrubbed base that never holds those vars — a successful eval always emits them, and the invariant holds unconditionally, not just "while the daemon runs outside any nix shell."

That said, the comment stated the invariant without the scrub, and the diff-omission hazard is real enough that you (reasonably) reconstructed the failure from it. 81d5fcd adds the grounding: the comment now names `env_clear()` to `SAFE_ENV_VARS` as the reason the diff can never omit the signature vars. Spec 03's "proves the flake evaluated" stays as-is — the proof is unconditional given the scrub, so the constraint qualifier would be wrong, not cautious.

No behavior change; the PATH-based discrimination would be redundant given the scrub (and `PATH` is forwarded, so a `/nix/store` PATH segment in the parent *would* survive into the diff base — the weaker signal, not the stronger).

amackillop: You are right, and I checked rather than took your word for it: `env_clear()` at cli_runner.rs:171, `safe_env()` at direnv.rs:340, and SAFE_ENV_VARS carrying neither signature var. The diff base is scrubbed, the emission is unconditional, and my failure scenario never reaches the subprocess. Withdrawn.

Your PATH point cuts the other way as I suggested it, too — PATH is on the safe list, so it is the one signal the parent *can* contaminate. My "stronger signal" was the weaker one.

81d5fcd naming the scrub in the comment is the right fix regardless: the invariant was true but unstated, and I reconstructed a false failure from exactly that gap. Spec staying unqualified is correct.

kitaebot: Appreciated — and the "checked rather than took your word for it" is the right standard; I'd rather be told BULLSHIT with the trace than agreed with politely. Thread closed, nothing further from my side.

PR #122, opened 2026-08-29
Plan/execute split: planner model thinks, flash default executes

### LINKED ISSUE #105: Plan/execute split: strong model writes the plan, cheap model drives the implementation loop
## Problem

Model choice is static per role (model_overrides), so a task pays strong-model rates for its entire span. The phases don't cost the same to get right: diagnosis and planning need the strong model; mechanical implementation — applying planned edits, running checks, iterating on compiler errors — mostly doesn't. Recent builds run $0.35–0.75/turn with ~2M prompt tokens/turn; the implementation tail is the bulk of those tokens.

## Proposal

Within-task phase transition. The strong model explores and produces the plan (through the first landed edit, so the pattern is established in context); the session then switches to a cheaper executor model that drives the loop for the remainder.

The whole idea collapses if the executor fumbles tool calls, so it's gated:

- **Entry gate**: switch only after a concrete plan artifact exists in context, and pin the plan across compaction so the executor never loses it.
- **Escalation back**: N consecutive tool-call failures, a policy strike, or a failed `just check` after M attempts hands the turn back to the strong model. One-way per phase — no ping-ponging.
- **Evidence before default-on**: measure the candidate executor's edit-call hit rate and task completion on a sample of real tasks first. Per-task cost/outcome exists (spec 27); #104 adds the cache split. Edit-call success and task success decouple — a model can land every edit call and still stall the task — so measure both.

Prerequisites: #102 (post-edit previews + no-op guard) materially raises a weaker model's edit hit rate; #103 gives the executor compiler deltas to iterate against instead of gate failures.

## Design forks (why needs-plan)

1. Switch trigger: plan artifact + first edit, or plan artifact alone.
2. Executor candidates and how the offline sample eval runs.
3. Whether the executor drives sub-agents too or only the root loop.
4. How escalation interacts with turn accounting (both phases bill the same task).

amackillop:
Closes #105.

The plan-first choreography already separates thinking from doing — plan turn, human gate, execution turn — but both phases ran on one root model, so the implementation bulk paid strong-model rates for work a gate-reviewed plan mostly de-risks. This splits them: plan turns think on a new `model_overrides.planner`, everything else rides the default, and the deploy config drops the default to `z-ai/glm-5.3-flash` with `glm-5.3` as the planner.

Five commits:

1. **Route plan turns to a planner model override** — `TurnRole` on the envelope, `send_message_with_role`, per-envelope provider selection in the actor with the ledger stamping the model that actually ran (so `/usage` separates the phases for the executor evaluation with no new columns), GitHub issues mapping `TurnKind::Plan` to it. Post-plan comment turns — revisions included — deliberately ride the default: the channel cannot tell approval from feedback pre-dispatch, and a revised plan still passes the plan gate.
2. **Embed the recorded plan in post-plan dispatches** — the execution turn must see the gate-reviewed plan text unconditionally. Session history is not relied on: post-compaction the assembled prompt holds a summary (LCM keeps the bytes but recovery is a tool round the model must initiate; flat is lossy outright). The lookup is pure — the comments were fetched this tick. This also removed the plan-pinning seam with #101.
3. **Executor-facing "Implementation notes" in the plan format** — the review brief now ends with a section addressed to the executing turn: files and seams per piece, verify commands, behaviors its tests must pin (so tests assert intent, not mirror implementation), ordering constraints, landmines. Pointers and constraints, never a script. Reviewers stop reading above the line; the plan gate reviews it with the rest.
4. **The deploy model split** — flash default, glm-5.3 planner (not kimi: the plan gate's judge must not be the author), plus the audit's one hard finding: `memory.model` pinned to glm-5.3, since distillation had only a reasoning bound and would otherwise inherit flash+low for facts that inject into every future turn. Duties, chat, and PR turns were audited and left on flash deliberately — their quality surfaces where a human sees it.
5. **Linear parity** — Linear's plan-first announcements also dispatch with the planner role; without this the flip would have silently degraded Linear planning to flash.

Rollout: everything is inert until the rebuild; rollback is one config line. The ledger's per-turn model stamp gives the flash-vs-glm-5.3 outcome and cost comparison on execution turns — that data decides whether the flash default survives.

### DIFF
--- a/specs/26-linear.md
+++ b/specs/26-linear.md
@@ -51,7 +51,10 @@ issues channel ([spec 25](25-github-issues.md)): an issue carrying
 `linear.plan_label` (default `needs-plan`, case-insensitive) gets
 plan-first; without it the announcement is a direct execution turn,
 with the same judgment backstop. The label is read at announcement
-time.
+time. Plan announcements dispatch with the planner turn role, exactly
+as on GitHub (spec 25's plan/execute split): both ticket channels'
+plans think on `model_overrides.planner`, everything else on the
+default.
 
 **Repo selection**: the target repository comes from a label on the issue —
 a label named like `owner/repo` (contains exactly one `/`). Issues without
--- a/specs/26-linear.md
+++ b/specs/26-linear.md
@@ -51,10 +51,13 @@ issues channel ([spec 25](25-github-issues.md)): an issue carrying
 `linear.plan_label` (default `needs-plan`, case-insensitive) gets
 plan-first; without it the announcement is a direct execution turn,
 with the same judgment backstop. The label is read at announcement
-time. Plan announcements dispatch with the planner turn role, exactly
-as on GitHub (spec 25's plan/execute split): both ticket channels'
-plans think on `model_overrides.planner`, everything else on the
-default.
+time. Plan announcements dispatch with the planner turn role, as on
+GitHub (spec 25's plan/execute split): both ticket channels' plans
+think on `model_overrides.planner`, everything else on the default.
+Routing parity only: GitHub's other half — the recorded plan comment
+embedded verbatim in post-plan dispatches — is not yet ported, so a
+Linear execution turn still relies on the plan surviving session
+context.
 
 **Repo selection**: the target repository comes from a label on the issue —
 a label named like `owner/repo` (contains exactly one `/`). Issues without
--- a/src/prompts/plan-format.md
+++ b/src/prompts/plan-format.md
@@ -33,9 +33,22 @@ matter, so omit it without comment:
   couple of pieces: independently shippable pieces in dependency
   order, a `mermaid` flowchart when the graph is not a straight line,
   and a note on what is inert until the last piece lands. Never a
-  commit-by-commit script: you will re-derive the commits cheaply at
-  execution time. Record sequencing detail only where re-deriving it
-  would be expensive — an ordering constraint you had to dig for.
+  commit-by-commit script; the commits are re-derived at execution
+  time. Record sequencing detail only where re-deriving it would be
+  expensive — an ordering constraint you had to dig for.
+- **Implementation notes** — always last, addressed to the executing
+  turn, which may run on a weaker model than the one planning:
+  reviewers can stop reading above this line. Pointers and
+  constraints, never a script — a step-by-step written before
+  touching the code is wrong in detail, and scripts get followed off
+  cliffs. Name the files and seams each piece touches, the command
+  that proves each piece works, the ordering constraints you dug for,
+  and the landmines you already spotted (a lint that will fire, a
+  test that must move, a fence that must not). For each piece, state
+  the behaviors its tests must pin — observable outcomes and the edge
+  cases you already found, never test code — so the tests assert what
+  the plan intends rather than mirroring whatever the implementation
+  happens to do.
 - **Kill switch / rollback** — only for changes that alter live
   behavior: the shortest way back.
 - **Out of scope (recorded, not built)** — only when you actually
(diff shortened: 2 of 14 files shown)


src/prompts/plan-format.md
```diff
@@ -33,9 +33,22 @@ matter, so omit it without comment:
   couple of pieces: independently shippable pieces in dependency
   order, a `mermaid` flowchart when the graph is not a straight line,
   and a note on what is inert until the last piece lands. Never a
-  commit-by-commit script: you will re-derive the commits cheaply at
-  execution time. Record sequencing detail only where re-deriving it
-  would be expensive — an ordering constraint you had to dig for.
+  commit-by-commit script; the commits are re-derived at execution
+  time. Record sequencing detail only where re-deriving it would be
+  expensive — an ordering constraint you had to dig for.
+- **Implementation notes** — always last, addressed to the executing
```
kitaebot: Placement contradicts the section's stated contract. This bullet sits under the header "The rest exist only when the task earns them … omit it without comment" and carries the conditional bullet style, but its own text says "always last" and both the commit message and spec 25 describe the section as mandatory. A plan for a small change — exactly the case where "What needs to be built" is omitted — leaves the prompt telling the planner this section may be dropped, defeating the executor-facing purpose it exists for. Move it into the "Every plan has these" group, or mark it explicitly always-present.

```suggestion
- **Implementation notes** — mandatory, always last, addressed to the executing
```

specs/26-linear.md
```diff
@@ -51,7 +51,10 @@ issues channel ([spec 25](25-github-issues.md)): an issue carrying
 `linear.plan_label` (default `needs-plan`, case-insensitive) gets
 plan-first; without it the announcement is a direct execution turn,
 with the same judgment backstop. The label is read at announcement
-time.
+time. Plan announcements dispatch with the planner turn role, exactly
+as on GitHub (spec 25's plan/execute split): both ticket channels'
```
kitaebot: "Exactly as on GitHub" overstates what was ported. Routing parity is real — plan announcements dispatch with the planner role — but Linear has no equivalent of the GitHub plan-embedding: no recorded plan-comment id, no plan body inlined in post-plan comment turns, no "Your current plan" in `format_comment`. The unconditional gate-reviewed-plan guarantee the PR's core commit establishes for GitHub does not exist on Linear, so a post-compaction Linear execution turn still relies on session history — precisely the failure mode the embedding was designed out for. Either scope this sentence to routing parity, or port the embedding.

kitaebot (approved):
Verdict: **correct** — the plan/execute split does what its commits claim: routing is per-envelope and consistent with the ledger stamp, the plan embedding is a pure in-memory lookup over comments already fetched this tick, and the config/spec/test wiring all line up with the behavior.

Verified across the five commits: `Agent::handle_message` selects the provider from `envelope.role` once and stamps the same provider's model into the ledger row, so `/usage` separates phases without new columns. GitHub maps only `TurnKind::Plan` to the planner role, and every other call site rides `send_message` → default — nothing was missed, including the deliberate decision to route post-plan comment turns (revisions included) to the default. The 100-comment fetch cap and deleted-plan cases degrade to today's id-only reference rather than failing the dispatch, and the new test pins both branches. `ModelOverrides::planner` is in `iter()`, so it inherits the same validation as the other overrides; an unset planner correctly collapses to the root provider. Linear gets an explicit role field rather than inferring from `needs_checkout`, which is the right call — correlation is not a contract.

Two non-blocking findings below. The first is a real one to fix before this lands: the "Implementation notes" bullet is placed under the "omit it without comment" header while its own text, the commit message, and spec 25 all call it mandatory — and the weak executor this PR enables is exactly the consumer that section exists for. The second: Linear got routing parity but not the plan-embedding guarantee, so scope the spec sentence or port the embedding.

amackillop:
Both findings addressed:

- Implementation-notes placement: fixed in b479efd — the section moved into the mandatory "every plan has these" list (its position in the written plan stays last), so small plans can no longer read the conditional header as permission to drop it.
- Spec 26 parity overstatement: fixed in 810160e — the sentence now claims routing parity only and names the missing embedding guarantee plainly. Porting the embedding to Linear (comment-id capture in the client, poll-state recording, embed from the fetched thread) is filed as #123.

PR #126, opened 2026-09-01
Block issue-comment writes through github_api

### LINKED ISSUE #116: Plan turns can double-post: github_api allows writing issue comments the channel already posts
## Problem

The plan-first choreography's contract is that the turn's final reply IS the plan: `plan-format.md` says "Your reply is posted verbatim as a comment on the ticket", and `dispatch()` (channel/github/issues.rs) posts it and records that comment's id in `plan_comments` as the anchor revision turns are pointed at.

On #114 the model broke the contract (daemon log, turn 11): right after the plan-gate reviewer returned findings, it posted the plan itself via `github_api POST issues/114/comments` (comment 5447407609, 01:52:16), then finished with a recap reply ("Plan posted to #114: ..."), which the channel posted verbatim as a second, redundant comment (5447409817, 01:52:37).

Two defects, the invisible one worse:

- The duplicate comment.
- `plan_comments["amackillop/kitaebot#114"]` now anchors to the recap's id, not the plan's. The next revision turn on #114 will be told the wrong comment is its plan.

The capability gap: `github_api` blocks non-GET only on `actions` and `dependabot` (api.rs `READONLY_RESOURCES`), so `POST issues/{n}/comments` sails through. Nothing in the codebase legitimately posts issue comments that way — the channels post replies, PR diff replies and comment edits have dedicated tools — so by the every-permission-needs-a-concrete-caller rule the write should not be reachable. The prompt side contributed: `review-gates.md` describes the plan lifecycle as "before posting it for sign-off", which reads as an action once a posting-capable tool exists.

## Proposal

1. **Mechanical block**: in `validate_path`, reject non-GET methods on `issues` paths containing a `comments` segment (covers `issues/{n}/comments` and `issues/comments/{id}`), with guidance that the channel posts the turn's final reply as the comment. Label writes (`issues/{n}/labels`) stay allowed. GETs stay allowed.
2. **Prompt clarification**: reword `review-gates.md`'s "posting it for sign-off" so the reply is unambiguously the posting; add one line to `plan-format.md`: never post the plan via a tool.

## One-time repair

`plan_comments` for `amackillop/kitaebot#114` should point at 5447407609 (the plan), not 5447409817 (the recap). Fix by editing the issues channel's persisted state in the VM's state db, or accept the degraded anchor — the recap does summarize the plan.

amackillop:
Closes #116.

On #114 the plan turn posted its plan through `github_api POST issues/114/comments` mid-turn, then replied with a recap — so the thread got the plan twice, and `plan_comments` anchored the issue to the recap instead of the plan. The write was reachable because only `actions` and `dependabot` are read-only resources, and nothing in the codebase legitimately writes issue comments through the escape hatch: the channels post replies, diff replies and comment edits have dedicated tools.

Two commits, suspenders then belt:

1. **Block issue-comment writes through `github_api`** — non-GET on `issues` paths containing a `comments` segment is refused (covers `issues/{n}/comments` and `issues/comments/{id}`); label writes and all GETs untouched. The guidance string tells the model where its comment actually goes, so the refusal teaches instead of just blocking.
2. **Make the prompts say the reply is the posting** — `review-gates.md`'s "before posting it for sign-off" framed posting as an action the model performs, which is the instruction that invited the tool write; it now says delivering the plan means ending the turn with it as the reply. `plan-format.md` states the same rule from the channel's side.

The #114 state itself was repaired at the time (plan anchor repointed to the real plan comment); this makes the recurrence structurally impossible.

### DIFF
--- a/src/tools/github/api.rs
+++ b/src/tools/github/api.rs
@@ -58,8 +58,9 @@ struct Args {
     method: Method,
     /// Path under `repos/<owner>/<repo>/`, e.g. `issues/42/comments`
     /// or `pulls?state=open`. Must start with one of: actions (GET
-    /// only), dependabot (GET only), issues, labels, milestones,
-    /// pulls, releases.
+    /// only), dependabot (GET only), issues (comment paths GET only —
+    /// the channel posts your final reply as the thread comment),
+    /// labels, milestones, pulls, releases.
     path: String,
     /// JSON body for POST/PATCH.
     #[serde(default, deserialize_with = "string_or_value")]
--- a/src/tools/github/api.rs
+++ b/src/tools/github/api.rs
@@ -113,6 +113,20 @@ fn validate_path(path: &str, method: Method) -> Result<(), ToolError> {
             guidance: format!("{resource} is read-only through this tool: use GET"),
         });
     }
+    // Covers `issues/{n}/comments` and `issues/comments/{id}`. The
+    // channels post the turn's reply as the comment; a comment written
+    // here duplicates it and steals the plan anchor (issue #116).
+    if resource == "issues"
+        && !matches!(method, Method::Get)
+        && path.split(['/', '?']).any(|segment| segment == "comments")
+    {
+        return Err(ToolError::Blocked {
+            operation: format!("github_api {} {path}", method.as_str()),
+            guidance: "issue comments are read-only through this tool: your final \
+                       reply is posted to the thread by the channel"
+                .into(),
+        });
+    }
     // '%' is blocked wholesale: percent-encoding can smuggle a '/'
     // past the segment check, and none of the allowed resources
     // need it.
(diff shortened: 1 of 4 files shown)


src/tools/github/api.rs
```diff
@@ -113,6 +113,20 @@ fn validate_path(path: &str, method: Method) -> Result<(), ToolError> {
             guidance: format!("{resource} is read-only through this tool: use GET"),
         });
     }
+    // Covers `issues/{n}/comments` and `issues/comments/{id}`. The
```
kitaebot, line 117: Nit: per AGENTS.md, rationale duplicating the commit message belongs in the commit message. The first line naming the two covered path shapes is the non-obvious fact worth keeping; the duplication/plan-anchor backstory is already in 45ab64f.

src/tools/github/api.rs
```diff
@@ -113,6 +113,20 @@ fn validate_path(path: &str, method: Method) -> Result<(), ToolError> {
...
+    // channels post the turn's reply as the comment; a comment written
+    // here duplicates it and steals the plan anchor (issue #116).
+    if resource == "issues"
+        && !matches!(method, Method::Get)
+        && path.split(['/', '?']).any(|segment| segment == "comments")
```
kitaebot, line 122: Nit: the any-segment match is broader than "comment writes" — it also blocks e.g. `issues/comments/{id}/reactions` and a DELETE of a label literally named `comments` via `issues/{n}/labels/comments`. Neither has a known legitimate caller, but the spec/guidance describe a narrower rule than the code enforces. Fine as-is if the over-block is intentional; worth a word in the spec if so.

kitaebot (approved):
Verdict: **correct** — the guard blocks exactly the issue-comment write endpoints before any network call, leaves label writes and GETs intact, and the tests cover the claimed shapes.

The mechanics check out: the new check sits in `validate_path` before the HTTP call, after the allowlist/readonly checks and before the `%` rejection (so percent-smuggled `comments` segments still fail); the any-segment match handles `issues/{n}/comments`, `issues/comments/{id}`, and query variants without false-positives on query strings; `pr_diff_reply`/`comment_update` bypass `validate_path` entirely, so no legitimate path breaks. Retargeting the three body-serialization tests to `issues/42/labels` was necessary and preserves what they test. Spec and prompt edits match the code.

One should-fix:

- **src/tools/github/api.rs:60** — the `path` arg's model-visible schema description still annotates only `actions`/`dependabot` as GET-only and lists `issues` as freely writable. The commit message frames the error guidance as "the fix for the next attempt", but the model only sees that after a refused call; annotating the new restriction in the schema (as is already done for actions/dependabot) prevents the attempt in the first place. That's the cheaper half of the belt-and-suspenders pair.

amackillop:
Finding addressed in 91369ba: the path parameter's schema description now annotates the issues comment-path restriction inline (matching the actions/dependabot GET-only annotations) and says where the reply actually goes, so the model reads the rule before calling instead of learning it from the refusal.

Issue #70, opened 2026-08-24
LCM externalization hands out absolute payload paths that file_read rejects


amackillop:
When LCM externalizes oversized tool output it tells the model the payload lives at an absolute path like /var/lib/kitaebot/context/lcm/payloads/file_<hash>. file_read rejects absolute paths, so the model's first (and correct-looking) move fails:

    ERROR ... Tool execution failed: Blocked: /var/lib/kitaebot/context/lcm/payloads/file_<hash> (absolute paths not allowed) tool=file_read

Six occurrences on 2026-08-23 between 18:45 and 19:04, in both the root turn and explore sub-agents, each costing an iteration plus a retry. The system pointed at a door and the doorman turned the model away.

The payload IS reachable via a workspace-relative path (the daemon's own memory note from 20:56Z confirms sub-agents recovered content that way), so this is purely a contract mismatch. Fix either side, one of:

- LCM emits the workspace-relative path in the externalization notice, or
- file_read accepts absolute paths under the LCM payloads root.

First option is smaller and keeps file_read's policy simple.

Evidence: journal 2026-08-23T18:45:21Z, 19:00:07Z, 19:04:04/06/13/15Z.

PR #130, opened 2026-09-01
Accept in-workspace absolute paths; make the diffs convention visible

### LINKED ISSUE #129: PathGuard refuses absolute paths under the workspace root it advertises
## Problem

The #126 review turn (2026-09-01 02:00) burned a tool round on four blocked reads in one shot: `/var/lib/kitaebot/memory/MEMORY.md`, `/var/lib/kitaebot/state/review-checklist.md`, and two review artifacts, all refused with "absolute paths not allowed". Every one is byte-identical to a workspace-relative path the guard would have served. This is the recurring "#70 family" failure: the model reaches for absolute paths, the tools refuse, iterations are lost.

The contradiction is ours: every agent prompt states "Working directory: /var/lib/kitaebot" (`compose_prompt`, and the root prompt equivalent), then `PathGuard::validate_and_join` rejects any absolute path — including ones under that very root. The rejection is not load-bearing for confinement: `ensure_under_root` canonicalize-checks the resolved path regardless, so accepting under-root absolutes by stripping the root prefix preserves the exact invariant.

## Proposal

In `validate_and_join`: an absolute path that `strip_prefix(workspace root)` succeeds on is rewritten to the remainder and flows through the existing checks (null byte, traversal, join, `ensure_under_root`). Absolute paths outside the root stay blocked, with guidance saying the path is outside the workspace rather than the current blanket refusal.

Prompt nudges ("use relative paths") have not extinguished this class and never will — models echo paths they see in `pwd` output, exec results, and their own prompts. Accepting the unambiguous form is the mechanical fix; the block remains for everything genuinely outside.

amackillop:
Closes #129.

The #126 review turn burned a tool round on four blocked `file_read`s, each an absolute spelling of a path the guard serves in relative form — the recurring "#70 family" failure. The contradiction was ours: every prompt advertises `Working directory: /var/lib/kitaebot`, models echo paths they are shown, and the guard refused the echo.

Two commits, same theme — stop hiding paths from the model and then punishing it for guessing:

1. **Accept absolute paths under the workspace root.** The blanket rejection was not load-bearing for confinement (`ensure_under_root` canonicalize-checks everything anyway), but it was load-bearing for the daemon-owned write fence, which compares components from the front and ran on the raw string: naive prefix-stripping would have let `/var/lib/kitaebot/state/x` writes walk past it with `var` as the first component. Normalization therefore lives in one function (`workspace_relative`) consumed by both the fence and the join, with a test pinning the bypass shut. Traversal is still checked on the original string; absolute paths outside the root stay blocked with guidance naming the actual problem.
2. **Rename `.diffs/` to `diffs/`.** The review-artifact convention was followed by some turns and freelanced by others (`projects/` dumps, an invented `.diff/run.sh`) — a dotdir is invisible in `ls`, so a model that didn't internalize the protocol never sees the prior art. 683d4cf's location rationale (workspace root, outside every checkout, never dirtying a tree being committed from) is preserved and now stated in spec 20 alongside the why-visible rationale. Existing `.diffs/` files wait for the workspace-hygiene sweep.

### DIFF
--- a/specs/15-sandbox.md
+++ b/specs/15-sandbox.md
@@ -63,7 +63,8 @@ The `exec` tier (`Policy::child_exec`):
 | Path | Access | Effect |
 |------|--------|--------|
 | Workspace root | list only (`ReadDir`) | Navigation works. Landlock rules are recursive, so `ReadFile` here would grant reads of everything beneath; with list-only, file reads and all writes under `state/`, `context/`, `memory/`, and `config.toml` are denied, and new root-level files cannot be created |
-| `projects/` | full | Builds, checkouts, `diffs/` all work |
+| `projects/` | full | Builds and checkouts work |
+| `diffs/` | full | Packed review diffs (spec 20): exec redirects write here. Pre-created by workspace init — the root is list-only for exec children, so only the daemon can create it, and Landlock cannot grant a missing path |
 | `state/review-checklist.md` | read | The one model-facing state file |
 | `context/lcm/payloads/` | read | The LCM payload store: `<file>` references hand these workspace-relative paths to the model ([spec 14](14-context-engine.md)), so shell reads of them must work. Optional — created lazily on first externalization, and references only exist after that point |
 | `.cache`, `.cargo`, `.npm`, `.local/{share,state}/pnpm` | full | `HOME` is the workspace root on the VM; nix flake eval fails hard without `~/.cache/nix`, and cargo/npm/pnpm need their caches. Provisioned by tmpfiles — Landlock cannot grant a path that does not exist. Grants are named, never speculative: onboarding a repo whose toolchain caches outside XDG (`.m2`, `.gradle`, `~/go/pkg/mod`, …) means adding its rule in `Policy::child_exec` plus a tmpfiles entry in the same commit |
--- a/specs/15-sandbox.md
+++ b/specs/15-sandbox.md
@@ -63,7 +63,7 @@ The `exec` tier (`Policy::child_exec`):
 | Path | Access | Effect |
 |------|--------|--------|
 | Workspace root | list only (`ReadDir`) | Navigation works. Landlock rules are recursive, so `ReadFile` here would grant reads of everything beneath; with list-only, file reads and all writes under `state/`, `context/`, `memory/`, and `config.toml` are denied, and new root-level files cannot be created |
-| `projects/` | full | Builds, checkouts, `.diffs/` all work |
+| `projects/` | full | Builds, checkouts, `diffs/` all work |
 | `state/review-checklist.md` | read | The one model-facing state file |
 | `context/lcm/payloads/` | read | The LCM payload store: `<file>` references hand these workspace-relative paths to the model ([spec 14](14-context-engine.md)), so shell reads of them must work. Optional — created lazily on first externalization, and references only exist after that point |
 | `.cache`, `.cargo`, `.npm`, `.local/{share,state}/pnpm` | full | `HOME` is the workspace root on the VM; nix flake eval fails hard without `~/.cache/nix`, and cargo/npm/pnpm need their caches. Provisioned by tmpfiles — Landlock cannot grant a path that does not exist. Grants are named, never speculative: onboarding a repo whose toolchain caches outside XDG (`.m2`, `.gradle`, `~/go/pkg/mod`, …) means adding its rule in `Policy::child_exec` plus a tmpfiles entry in the same commit |
--- a/src/tools/path.rs
+++ b/src/tools/path.rs
@@ -62,13 +62,11 @@ impl PathGuard {
         self.resolve_new(path)
     }
 
-    /// The workspace-relative form of `path`. An absolute path under
-    /// the root names the same file as its relative spelling — the
-    /// prompt advertises the root, so models echo it (#129) — and is
-    /// stripped to that spelling; other absolute paths are refused.
-    /// Every consumer (the daemon-owned write fence, the join) must
-    /// see this form, or an absolute spelling would sidestep
-    /// component-based checks.
+    /// The workspace-relative form of `path`: an absolute path under
+    /// the root is stripped to the relative spelling it names; other
+    /// absolute paths are refused. Every consumer (the daemon-owned
+    /// write fence, the join) must see this form, or an absolute
+    /// spelling would sidestep component-based checks.
     fn workspace_relative<'a>(&self, path: &'a str) -> Result<&'a Path, ToolError> {
         match Path::new(path).strip_prefix(&self.root) {
             Ok(stripped) => Ok(stripped),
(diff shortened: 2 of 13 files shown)


specs/15-sandbox.md
```diff
@@ -63,7 +63,7 @@ The `exec` tier (`Policy::child_exec`):
 | Path | Access | Effect |
 |------|--------|--------|
 | Workspace root | list only (`ReadDir`) | Navigation works. Landlock rules are recursive, so `ReadFile` here would grant reads of everything beneath; with list-only, file reads and all writes under `state/`, `context/`, `memory/`, and `config.toml` are denied, and new root-level files cannot be created |
-| `projects/` | full | Builds, checkouts, `.diffs/` all work |
+| `projects/` | full | Builds, checkouts, `diffs/` all work |
```
kitaebot: Tier::Exec grants the workspace root ReadDir only, with no `diffs/` rule, so the prompt-documented `mkdir -p diffs && git diff ... > diffs/pr-<n>.diff` redirect fails with EACCES exactly as `.diffs/` did. That makes commit 2's closing claim — "new turns write diffs/ from the next rebuild" — unsupported by the mechanism, and this spec row now implies the `projects/` grant covers a directory that is not under `projects/`.

Either add a named `diffs/` grant in `child_exec` (plus the corresponding spec 15 table row) or correct the claim.

src/tools/path.rs
```diff
@@ -50,16 +52,36 @@ impl PathGuard {
...
     }
 
     /// Like [`PathGuard::resolve_new`], for tools that write the file.
     pub fn resolve_writable_new(&self, path: &str) -> Result<PathBuf, ToolError> {
-        deny_daemon_owned(path)?;
+        deny_daemon_owned(self.workspace_relative(path)?)?;
         self.resolve_new(path)
     }
 
+    /// The workspace-relative form of `path`. An absolute path under
```
kitaebot: This doc comment carries three lines of #129 backstory that the commit message already states; AGENTS.md sends rationale to the commit message. Keep the load-bearing invariant (every consumer must see the normalized form) and drop the history.

kitaebot (commented):
Verdict: **incorrect** — the path normalization itself is correct and the write-fence bypass is genuinely pinned shut, but the rename is incomplete and the absolute-path acceptance stops one tool short of its own goal.

**Summary**

Commit 1 is solid. `workspace_relative` normalizes via `Path::strip_prefix`, which is component-wise, so a sibling root like `/var/lib/kitaebot-evil` fails the strip and lands in the absolute-outside error — no prefix confusion. Traversal is still checked on the original string in `validate_and_join` before the join, pinned by the new test, and both the write fence (`resolve_writable`/`resolve_writable_new`) and the join consume the normalized form; grep confirms no raw-string consumer is left. Root-equality strips to an empty path and joins back to the root; harmless. `PathGuard::new` canonicalizes, so the advertised root and the stored root agree.

Findings:

1. **must-fix — `src/backup.rs:33`** (file not in the diff, so no inline anchor): the PR renames `.diffs/` → `diffs/` in spec 05's Durable/Derived table, but `DERIVED` — the code implementing that exact split — still lists only `".diffs"`. `report_unclassified` will warn "diffs is neither backed up nor classified derived" on every backup run, a false drift alarm from the mechanism whose design is that drift surfaces at every backup run. Add `"diffs"` alongside `".diffs"` (keep the legacy entry for the deferred hygiene sweep).

2. **should-fix — `src/tools/exec.rs:765`** (file not in the diff, so no inline anchor): `resolve_working_dir` still rejects absolute paths with "absolute paths not allowed". The PR's own rationale — prompts advertise `Working directory: /var/lib/kitaebot`, models echo it, prompt nudges don't extinguish the class — applies verbatim to `working_dir`, which models are told to set constantly. An echoed absolute `working_dir` still burns a turn on the same #70-family failure this PR exists to end. Accept the in-workspace absolute spelling here too (or share the normalization helper).

3. **should-fix — `specs/15-sandbox.md:66`:** Tier::Exec grants the workspace root ReadDir only and has no `diffs/` rule, so the prompt-documented `mkdir -p diffs && git diff ... > diffs/pr-<n>.diff` redirect fails with EACCES exactly as `.diffs/` did. Commit 2's "new turns write diffs/ from the next rebuild" is not supported by the mechanism, and the edited spec 15 row implies the `projects/` grant covers a directory that isn't under `projects/`. Either add a named `diffs/` grant in `child_exec` (with the spec 15 table row, per that spec's named-grants rule) or correct the claim.

4. nit — `src/tools/path.rs:65`: the `workspace_relative` doc comment carries three lines of #129 backstory the commit message already states; keep the load-bearing invariant, drop the history.

Everything else checks out: the `prs.rs` test needle matches the updated `review-protocol.md`; specs 05/15/20/23 and both prompts are internally consistent with the rename (the remaining `.diffs/` mention in spec 20 is the deliberate historical reference); and the commit messages' central claims — single normalization point, traversal on the original string, bypass pinned — are borne out by the code.

[ledger: finding ids 160, 161, 162, 163]

amackillop:
All four findings addressed:

- **162** (fixed, d15f136): workspace init pre-creates `diffs/` (the root is list-only for exec children, so only the daemon can) and the exec tier grants it by name, Optional per the child-tier invariant its own contract test enforces. Your finding also reframes the drift diagnosis: redirects into `.diffs/` have failed with EACCES since the tier tightening — the documented location was denied and `projects/` was the writable place that worked. The commit message records that.
- **160** (fixed, 93f9e3d): `DERIVED` gains `diffs` via the new workspace const; legacy `.diffs` entry kept for deployed workspaces until the hygiene sweep.
- **161** (fixed, 27fa3e1): `resolve_working_dir` accepts the in-workspace absolute spelling with the same normalization order (traversal on the original string first), tested both ways.
- **163** (fixed, b2dd00a): doc trimmed to the invariant.

kitaebot (approved):
Verdict: **correct** — the delta addresses all four findings, verified against the code.

- d15f136 — #162 fixed: `Workspace::init_at` pre-creates `diffs/` (idempotent via `create_dir_all`), the exec tier grants it by name as `Optional`, which is exactly what the `child_only_workspace_root_is_required` contract test enforces; spec 15 gets its row.
- 93f9e3d — #160 fixed: `crate::workspace::DIFFS_DIR` sits in `DERIVED`, the legacy `".diffs"` entry survives, and `stage_covers_every_durable_file_and_skips_derived` now pins the classification since `init_at` creates the dir.
- 27fa3e1 — #161 fixed: traversal still checked on the original string before any stripping, and the post-join `starts_with` guard closes the component-mismatch edge (e.g. `/workspace-differing/...`); tested both directions.
- b2dd00a — #163 fixed: doc comment trimmed to the invariant.

One nit on d15f136's commit message: it says redirects into `.diffs/` "have failed with EACCES since the tier tightening", but Landlock raises ENOENT when a `path_beneath` rule names a missing path, and `.diffs` was never granted by name — the failure's fatal step is misattributed. Code is right; the narrative should name the real errno or stay mechanism-neutral. Not blocking.

[ledger: finding ids 169]

PR #132, opened 2026-09-01
Pretty-print JSON payloads at externalization

### LINKED ISSUE #119: Single-line JSON LCM payloads defeat grep: every match returns the whole file
## Symptom

Turn 6 (Duty, `CumuloGlobal/lightning-node`, 2026-08-28T18:09Z) spent ~55 minutes across ~9 explore subagents trying to extract dependabot alert numbers from one LCM payload.

The payload `context/lcm/payloads/file_472d3e1474ac29c0` is 163 KB of single-line JSON (an externalized GitHub API response). grep is line-based, so any matching pattern returns the entire file:

```
18:14:41 tools::grep: pattern="number":\d+ path=context/lcm/payloads/file_472d3e1474ac29c0
18:14:41 tools: Tool finished tool=grep elapsed=1.8ms bytes=163698
18:15:19 tools::grep: pattern=/dependabot/alerts/(\d+) ...
18:15:19 tools: Tool finished tool=grep elapsed=1.8ms bytes=163698
```

Every "targeted" grep costs ~41k tokens. One explore subagent's prompt grew 3k → 135k in six iterations of re-receiving the same file. The parent kept spawning fresh explores that rediscovered the same trap, one after another, from 18:13 to 19:07.

`lcm_grep` in fts mode is no escape hatch either: `"number":750` fails FTS5 parsing (`no such column: number`), hit by at least two separate agents in the same window.

## Fix options

1. Pretty-print JSON before externalizing to an LCM payload, so line-based tools produce line-sized results. Cheap, fixes grep, file_read ranges, and lcm_grep regex mode all at once.
2. And/or make grep truncate a matched line around the match site instead of returning it whole.

Option 1 is the structural fix: the payload store exists so agents can retrieve *parts* of large outputs, and a single-line blob makes every partial-retrieval tool degenerate to full-file reads.

kitaebot:
Closes #119.

The LCM engine now pretty-prints valid JSON payloads before writing the disk copy under `context/lcm/payloads/`, so grep, `file_read` ranges, and `lcm_grep` regex mode all produce line-sized results instead of re-reading a 163KB one-line blob per match.

- `explore::normalize_payload` (pure): parse → `to_string_pretty`; non-JSON payloads stay byte-identical (`Cow::Borrowed` fast path). A line-numbered `file_read` result is pretty-printed underneath its `<tool_output>` framing with fresh sequential numbers (file_read parity); any other framing is dropped so the body is plain pretty JSON. Round-trip fidelity caveats (key re-sorting, >64-bit ints → f64) are noted at the site and are harmless for API payloads.
- `externalize` writes the normalized bytes; `file_id`, `byte_size`, `token_count` stay derived from the raw content, so content-addressed dedupe and context accounting are unchanged.
- `lcm_expand` recovery reads the same on-disk copy and is byte-identity agnostic.

Validation: 5 new `normalize_payload` unit tests plus 2 updated engine tests; full `nix flake check` green (cargo-deny duplicate-crate warnings pre-existing). The in-sandbox `just rust-check` failure set is identical to the clean-master baseline (32 known gpg/git-fixture failures, confirmed via stash-compare).

Note: this fixes new payloads only. Existing stored blobs keep their shape (ids are content-addressed, so re-externalizing the same content dedupes to the old file) — acceptable, since old payloads age out with conversations.

### DIFF
--- a/src/context/lcm/explore.rs
+++ b/src/context/lcm/explore.rs
@@ -113,34 +113,43 @@ pub fn normalize_payload(content: &str) -> std::borrow::Cow<'_, str> {
     };
     // Round-trip fidelity: object keys are re-sorted (Map is a
     // BTreeMap by default) and >64-bit integers approximate to f64;
-    // harmless for API payloads. Serialization of a parsed Value
-    // cannot fail, so fall back to verbatim on the impossible.
-    let pretty = serde_json::to_string_pretty(&parsed).unwrap_or_else(|_| content.to_string());
-    // Re-emit the framing (with fresh sequential numbers) only for
-    // line-numbered file_read results, keeping file_read parity; any
-    // other framing is dropped so the stored body stays a plain
-    // pretty JSON.
-    let Some((open, close)) = framed_numbering(content) else {
+    // harmless for API payloads.
+    let Ok(pretty) = serde_json::to_string_pretty(&parsed) else {
+        // Unreachable: serializing a parsed Value cannot fail. Return
+        // the original rather than wrap garbage in fresh framing.
+        return std::borrow::Cow::Borrowed(content);
+    };
+    // Re-emit the framing only for line-numbered file_read results,
+    // keeping file_read parity; any other framing is dropped so the
+    // stored body stays a plain pretty JSON.
+    let Some(open) = framed_numbering(content) else {
         return std::borrow::Cow::Owned(pretty);
     };
     let mut rebuilt = String::with_capacity(pretty.len() + (content.len() - unwrapped.len()));
     rebuilt.push_str(open);
-    if !open.is_empty() {
-        rebuilt.push('\n');
-    }
+    rebuilt.push('\n');
     for (i, line) in pretty.lines().enumerate() {
         let _ = writeln!(rebuilt, "{}\t{line}", i + 1);
     }
-    rebuilt.push_str(close);
+    // Fresh stats over the stored body: strip_line_numbering requires
+    // the trailer, so omitting it would leave invalid file_read
+    // framing.
+    let lines = pretty.lines().count();
+    let _ = writeln!(
+        rebuilt,
+        "({lines} lines shown, {lines} total, {} bytes)",
+        pretty.len()
+    );
+    rebuilt.push_str("</tool_output>");
     std::borrow::Cow::Owned(rebuilt)
 }
 
 /// The framing of a line-numbered `file_read` result: the
-/// `<tool_output>` open line and its close tag, re-emitted around the
-/// pretty body so the stored copy keeps `file_read` parity. Unnumbered
-/// framing is not rebuilt — the body gets no line numbers of its own,
-/// so re-wrapping would wrap a bare pretty JSON and inject nothing.
-fn framed_numbering(content: &str) -> Option<(&str, &str)> {
+/// `<tool_output>` open line, re-emitted around the pretty body so
+/// the stored copy keeps `file_read` parity. Unnumbered framing is
+/// not rebuilt — the body gets no line numbers of its own, so
+/// re-wrapping would wrap a bare pretty JSON and inject nothing.
+fn framed_numbering(content: &str) -> Option<&str> {
     let (first, rest) = content.split_once('\n')?;
     if !TOOL_OUTPUT_OPEN_RE.is_match(first) {
         return None;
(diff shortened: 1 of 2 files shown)


src/context/lcm/explore.rs
```diff
@@ -99,6 +100,59 @@ pub fn format_file_reference(
...
+    let Some((open, close)) = framed_numbering(content) else {
+        return std::borrow::Cow::Owned(pretty);
+    };
+    let mut rebuilt = String::with_capacity(pretty.len() + (content.len() - unwrapped.len()));
+    rebuilt.push_str(open);
+    if !open.is_empty() {
+        rebuilt.push('\n');
+    }
+    for (i, line) in pretty.lines().enumerate() {
+        let _ = writeln!(rebuilt, "{}\t{line}", i + 1);
```
amackillop, line 132: The rebuild drops the `(N lines shown, M total, B bytes)` trailer, so the stored copy is no longer valid file_read framing by this file's own definition — `strip_line_numbering` requires the trailer and will refuse to strip it. Nothing re-strips the disk copy today, so it's latent, but "file_read parity" in the comments oversells what's stored. Either re-emit a correct trailer over the pretty lines or call it partial framing honestly. (Also cosmetic: `writeln!`'s trailing newline plus the `"\n</tool_output>"` close leaves a blank line where the trailer used to sit.)

src/context/lcm/explore.rs
```diff
@@ -99,6 +100,59 @@ pub fn format_file_reference(
...
+    // Re-emit the framing (with fresh sequential numbers) only for
+    // line-numbered file_read results, keeping file_read parity; any
+    // other framing is dropped so the stored body stays a plain
+    // pretty JSON.
+    let Some((open, close)) = framed_numbering(content) else {
+        return std::borrow::Cow::Owned(pretty);
+    };
+    let mut rebuilt = String::with_capacity(pretty.len() + (content.len() - unwrapped.len()));
+    rebuilt.push_str(open);
+    if !open.is_empty() {
```
amackillop: Dead check: `open` matched `TOOL_OUTPUT_OPEN_RE`, which cannot match an empty line. Push the newline unconditionally.

src/context/lcm/explore.rs
```diff
@@ -99,6 +100,59 @@ pub fn format_file_reference(
...
+pub fn normalize_payload(content: &str) -> std::borrow::Cow<'_, str> {
+    let unwrapped = strip_tool_framing(content);
+    let Ok(parsed) = serde_json::from_str::<serde_json::Value>(&unwrapped) else {
+        return std::borrow::Cow::Borrowed(content);
+    };
+    // Round-trip fidelity: object keys are re-sorted (Map is a
+    // BTreeMap by default) and >64-bit integers approximate to f64;
+    // harmless for API payloads. Serialization of a parsed Value
+    // cannot fail, so fall back to verbatim on the impossible.
+    let pretty = serde_json::to_string_pretty(&parsed).unwrap_or_else(|_| content.to_string());
```
amackillop: The fallback is shape-wrong: if it ever fired, `pretty` would be the full framed original, and `framed_numbering(content)` below would then wrap framing around framing. It's unreachable (serializing a parsed `Value` to a `String` can't fail), but an unreachable branch that produces garbage is worse than one that returns `Borrowed(content)` early.

amackillop (commented):
Three nits from a pass over the diff — none blocking. The seams that could regress (expand accounting, the parse gate in front of the framing rebuild) both check out: `append_payload` re-estimates tokens from the actual on-disk bytes, and no line-numbered text can be valid JSON, so `framed_numbering`'s laxer check can't disagree with `strip_line_numbering` on anything that reaches it.

src/context/lcm/explore.rs
kitaebot: All three fixed in 862a88a: the fallback now returns the original early via let-else, the dead emptiness check is gone (newline pushed unconditionally, `framed_numbering` reduced to the open line), and the rebuild re-emits a fresh `(N lines shown, M total, B bytes)` trailer so the stored copy round-trips through `strip_tool_framing` â the test now asserts the trailer and pins that round-trip.

src/context/lcm/explore.rs
kitaebot: Fixed in 862a88a: the fallback returns `Cow::Borrowed(content)` early (as let-else, per clippy) instead of handing the framed original back into the framing rebuild.

src/context/lcm/explore.rs
kitaebot: Fixed in 862a88a: newline pushed unconditionally, and `framed_numbering` now returns just the open line instead of an (open, close) tuple.

amackillop (approved):

Issue #128, opened 2026-09-01
Failed provider attempts drop their billed usage from the ledger


amackillop:
## Problem

A failed provider attempt returns `Err(ProviderError)` with no usage attached, so tokens the provider billed for the failed draw never reach `TurnUsage` or the ledger. The 2026-09-01 01:55 empty on issue #121's turn is concrete: `finish_reason="error"` after 3317 reasoning tokens — billed by OpenRouter, invisible in `/usage`.

Pre-existing for every transient retry (a 5xx after partial generation, a malformed tool call whose draw was billed), and slightly amplified by #127 making empty responses retryable: each redraw is a billed request whose cost the ledger no longer sees. It is also an errors-never-discard-information violation in miniature — an empty response arrives with a perfectly good `usage` block that `parse_response` throws away with the parse failure.

## Proposal

Keep the usage when the parse fails: `attempt()` already builds `CallUsage` before parsing, so the parse-failure path can carry it — either a `usage: Option<CallUsage>` field on the parse-shaped `ProviderError` variants (`EmptyResponse`, `Truncated`, `MalformedToolCall`), or a small `ChatError { error, usage }` wrapper at the retry boundary that `chat()` accumulates from before surfacing the final error. The accumulated failed-attempt usage folds into the turn's `TurnUsage` alongside the successful call, so the ledger records what was actually billed rather than what happened to succeed.

Transport-level failures (`Network`) genuinely have no usage to keep; nothing to do there.

Sizing note: at observed rates (6 empties in 4 days, a few thousand tokens each) this is cents — the reason to fix it is that the ledger's contract is honest accounting, and per-turn cost comparisons (flash executor eval, spec 27) are quietly flattered by dropping the cost of exactly the failures the cheap models produce more of.

PR #137, opened 2026-09-01
Verify index cap post-pass with one compaction-only retry

### LINKED ISSUE #121: Memory index compaction does not converge: soft directive, no verification, races with main-agent writes
## Symptom (2026-08-28, 12:44–16:49 UTC)

```
WARN memory index still over cap after distillation: 8338B against a 8192B cap   (12:44)
WARN memory index exceeds cap, truncating cap_bytes=8192                          (12:50)
WARN memory index exceeds cap, truncating cap_bytes=8192                          (13:00)
WARN memory index still over cap after distillation: 8257B against a 8192B cap   (15:47)
WARN memory index still over cap after distillation: 8283B against a 8192B cap   (16:49)
```

Three consecutive distill passes failed to bring `memory/MEMORY.md` under its own 8192B cap, over 4+ hours. During that window every turn's injected memory was silently truncated at the byte cap (the cut fell in the Tickets section — the newest facts). The index is now 7561B; it converged by something other than the designed mechanism.

## Analysis

**The compaction directive is a soft side quest inside a pass whose primary job adds bytes.** `src/memory/distill.rs:237-246` injects "REQUIRED FIRST: compact the index" into a pass whose main purpose is folding new transcript facts *into* memory. When the fold outweighs the compaction, the pass ends over cap — and the post-pass check (`distill.rs:291-297`) only warns and tells the next pass the same thing. Same directive, same model, same result: three failures in a row. Nothing verifies the compaction happened, nothing escalates, nothing enforces.

**Injection truncates silently.** `src/memory/mod.rs:44-49` cuts at the cap on a UTF-8 boundary with an in-segment marker. The agent has no durable signal that half its memory was invisible for hours.

**Concurrent-writer race.** The main agent and the distiller both write `memory/MEMORY.md` and `memory/topics/*`. In the same window, four main-agent `file_edit` calls failed with "no match found... the file may have changed since you read it" (21:14, 23:16 on 08-27; 09:37, 15:46 on 08-28) — the distiller rewrote the files underneath. Each failure costs a retry iteration, and a main-agent write landing right after a successful compaction re-grows the index past the cap again.

## Suspected fix directions (need design)

- Verify post-pass: if the index is still over cap after the fold, run a compaction-only turn (or re-run with the directive as the sole task) before declaring the pass done — a bounded fixpoint instead of "required next pass".
- Or a mechanical fallback the LLM cannot fail: move the oldest finished-ticket topic sections to `memory/topics/` and leave pointers, deterministically, when the soft path fails once.
- Surface the truncation to the agent turn, not just the log, so the data loss is at least visible.
- The concurrent-writer race is arguably its own issue (the file_edit retry pattern papers over it), but any compaction fix has to answer what happens when the main agent writes between the distiller's read and write.

## Evidence locations

- Directive injection: `src/memory/distill.rs:237`
- Post-pass warn only: `src/memory/distill.rs:291-297`
- Silent truncation: `src/memory/mod.rs:44-49`
- Distiller tool set (file_write/file_edit on the same files the main agent edits): `src/memory/distill.rs:90-96`

kitaebot:
Closes #121 (plan approved: option A, bounded LLM retry).

Three consecutive distill passes left `memory/MEMORY.md` over its 8192-byte injection cap because compaction is a soft side quest inside a pass whose primary job — folding new facts — *adds* bytes. Two changes, one commit:

**Post-pass verification with one compaction-only retry.** After a successful fold, if the index is still over cap, the pass runs exactly one additional ephemeral turn whose only subject is the index (no transcript spans). The retry runs after the watermarks advance, so it is strictly best-effort: a failed retry logs a warning and flags the pass summary, never fails the pass — transcript work that succeeded stays done. Both turns' billed usage merges into one ledger row via a new `TurnUsage::add_turn`; the row's timing/outcome stay the main turn's.

**Truncation stops being anonymous.** The injection marker now names the `##` sections that fell wholly past the cut (`sections not shown: …`), so the invisible tail is visible as invisible — the file on disk is never truncated, only the injected copy.

Bounded by construction: exactly one retry (hardcoded, no config knob), injection always caps, and a double failure is today's terminal state plus named casualties. Spec 21 rows updated; distill prompt hardened ("folding facts into an invisible index writes into a void").

Validation: 5 new/updated distill run tests (retry fires over cap, skipped under cap, failed retry does not fail the pass, retry message is index-only, double-failure marker) plus 4 memory-injection tests (marker naming, straddling-section honesty, end-to-end truncation). `nix flake check` green.

Out of scope (recorded): the main-agent/distiller lost-update race on full `file_write` rewrites — deserves its own issue with its own design (versioning or single-writer).


amackillop (commented):
Two real findings and two minor ones. Finding 1 contradicts the spec text this PR adds and is worth fixing before merge; the rest holds up — retry ordering after `state.save` is correct, the double-failure check re-reads the file after the retry, `add_turn`'s None-handling mirrors `add_call`, and the `build_user_message` refactor is order-preserving.

src/memory/distill.rs
```diff
@@ -289,15 +305,107 @@ pub async fn run<P: Provider, E: ContextEngine>(
...
+/// folding pass keeps failing; a turn whose only subject is the index
+/// is the task the model can actually finish. Best-effort by design:
+/// the caller advances watermarks first, so any failure here just
+/// logs and returns an empty meter to fold into the pass's billing.
+async fn retry_compaction<P: Provider>(
+    workspace: &Workspace,
+    distiller: &Distiller,
+    provider: &P,
+    summarize: &SummarizeFn,
+    ephemeral: &mut EphemeralSession,
```
amackillop: **The retry turn is not index-only — it carries the whole fold turn.** This reuses the fold's `EphemeralSession`, which is an append-only `Vec<Message>` that never resets. The retry request therefore contains the fold's entire conversation: the transcript-spans user message, every tool call, the lot. The spec text added in this PR says "the index as the sole subject, no transcript spans", and the design rationale is that compaction fails when attention is diluted by transcripts — this recreates exactly that dilution, and re-bills the full fold context as prompt tokens on the retry.

The test doesn't catch it because it only asserts the *new user message* lacks "session history", not that the request does; the earlier fold message with all the spans is sitting right there in `last_request()`.

Fix is one line: build a fresh `EphemeralSession::new(DISTILL_TOOL_OUTPUT_TOKENS)` inside `retry_compaction`. If riding the fold's context was intentional (prompt-cache reasons, say), then the spec, PR body, and test comment are all wrong instead — either way something has to change.

src/memory/mod.rs
```diff
@@ -59,6 +61,30 @@ fn truncate_on_boundary(s: &str, cap: usize) -> (&str, bool) {
...
+/// is lost; the marker says the list covers whole sections.
+fn truncation_marker(full: &str, kept: &str, cap_bytes: usize) -> String {
+    let shown: std::collections::HashSet<&str> = trimmed_sections(kept).collect();
+    let dropped: Vec<&str> = trimmed_sections(full)
+        .filter(|title| !shown.contains(title))
+        .collect();
+    let mut marker = format!("[memory index truncated at {cap_bytes} bytes");
+    if !dropped.is_empty() {
+        marker.push_str(" — sections not shown: ");
+        marker.push_str(&dropped.join("; "));
```
amackillop: **The marker is unbounded, breaking the cap contract it sits next to.** Every dropped `##` title is joined into the marker, and the marker is appended *after* the cap cut. The spec paragraph this PR edits says the cap is the token contract "regardless of what the model wrote" — but a pathological MEMORY.md with hundreds of sections now produces an arbitrarily large marker injected past the cap. The old constant-size marker honored the contract; this one doesn't. Cap the list: first N titles plus "and K more".

src/memory/mod.rs
```diff
@@ -59,6 +61,30 @@ fn truncate_on_boundary(s: &str, cap: usize) -> (&str, bool) {
     (kept, kept.len() < s.len())
 }
 
+/// Why the tail is invisible and what it held: the marker names the
+/// `##` sections that fall entirely past the cut, so the agent (and
+/// the log) can see exactly what memory was hidden. A section
+/// straddling the cut keeps its header visible, so only its body tail
+/// is lost; the marker says the list covers whole sections.
+fn truncation_marker(full: &str, kept: &str, cap_bytes: usize) -> String {
+    let shown: std::collections::HashSet<&str> = trimmed_sections(kept).collect();
```
amackillop: Minor: filtering dropped sections by title against a `HashSet` of shown titles means a dropped `## Repos` vanishes from the marker if another `## Repos` survived the cut. Probably fine for an index the distiller writes, but the marker's whole job is honesty about what's hidden.

src/memory/distill.rs
```diff
@@ -289,15 +305,107 @@ pub async fn run<P: Provider, E: ContextEngine>(
...
+) -> TurnMeter {
+    let Some(directive) = index_over_cap(workspace, distiller.index_cap_bytes) else {
+        return TurnMeter {
+            usage: TurnUsage::default(),
+            started_at: crate::time::now_epoch(),
+            duration: std::time::Duration::ZERO,
+            outcome: "skipped",
```
amackillop: Cosmetic: this fabricated "skipped" meter (with a `started_at` the caller always overwrites) is a dummy built to satisfy the merge. `Option<TurnMeter>` would say what it means.

amackillop:
Conflict with master resolved (both sides appended a TurnUsage method; kept add_turn and #128's add_failed) and all four findings fixed in the follow-up commit:

1. **Retry rides no fold context**: retry_compaction builds its own EphemeralSession; the test now asserts the whole request carries nothing of the fold, not just the new message.
2. **Marker bounded**: at most 8 dropped titles plus '(+K more)'; the cap contract holds.
3. **Positional dropped-detection**: byte offsets instead of a title set — a surviving duplicate title cannot hide its dropped twin, and a header split mid-line by the cut counts as dropped (it is not shown whole). Both cases regression-tested.
4. **Option<TurnMeter>** replaces the fabricated 'skipped' meter; the ledger row starts from the main meter and folds the retry in only when it ran.

One cross-branch note: PR #150 (in flight) gives the distiller's fold turn a ToolCtx carrying its 20k inline cap; the retry turn added here should get the same ctx once both merge — one-line follow-up for whichever lands second.

src/memory/distill.rs
kitaebot: Verified against 427d613 — all four stand:

1. Confirmed: `retry_compaction` builds its own `EphemeralSession` (src/memory/distill.rs:333), and `compaction_retry_runs_when_the_pass_leaves_the_index_over_cap` asserts the whole captured request carries nothing of the fold, not just the new message.
2. Confirmed: `MARKER_SECTIONS_MAX = 8` bounds the list with "(+K more)", regression-tested by `truncation_marker_caps_the_dropped_section_list`.
3. Confirmed: detection is positional (byte offsets vs `kept.len()`), and `truncation_marker_lists_a_dropped_duplicate_title` covers the duplicate-title case. However, the claimed split-header regression test did not exist — added it in fe8fa67 (`truncation_marker_names_a_split_header`: kept ends `## To`, full has `## Tooling`, marker must name Tooling). Test-only; full gate green.
4. Confirmed: `retry_compaction` returns `Option<TurnMeter>` (`None` = under cap, no retry), and the ledger row folds the retry meter in only when it ran.

Noted the #150 cross-branch follow-up for the retry turn's ToolCtx.

src/memory/mod.rs
kitaebot: Confirmed fixed in 427d613: the list is capped at 8 titles (`MARKER_SECTIONS_MAX`) plus "(+K more)", so the marker cannot breach the cap contract it documents; regression-tested.

src/memory/mod.rs
kitaebot: Confirmed fixed in 427d613: detection is positional now, so a surviving duplicate cannot hide its dropped twin — regression-tested in `truncation_marker_lists_a_dropped_duplicate_title`. Your point also exposed that the claimed split-header regression test never existed; added it in fe8fa67.

src/memory/distill.rs
kitaebot: Confirmed fixed in 427d613: `retry_compaction` returns `Option<TurnMeter>`, `None` when the index is under cap and no retry ran; the ledger row folds the retry meter in only when it exists.

Issue #136, opened 2026-09-01
Deny-list guidance: 63 of 92 rules say only 'command blocked by policy'


amackillop:
The deny list supports per-rule guidance and the rules that use it work — the git rules' "use the git_fetch tool" steers retries correctly in the logs. But 63 of 92 rules ship the bare default, and the 2026-09-01 logs show what that costs on the flash default: a blocked `cat /tmp/...` was retried with "Please advise how to proceed." appended to the command, and two turns ground into POLICY_ROUND_LIMIT halts. The blocked model asked for exactly what a guidance string is.

Pass over the bare rules and write what-to-do-instead for every class with an obvious remediation: /tmp file references → the workspace paths file tools can read (the #70 family); raw cargo → the just recipes; state/context/memory writes → why they are daemon-owned and what the reviewer-checklist exception is; recursive rm → single-file rm is allowed; nix operations → the daemon never runs nix mutations. Rules with no sensible remediation (reverse shells, secret harvesting) keep the bare message — guidance there would be a how-to.

Add one new rule in the same pass, concrete caller in the 2026-09-01 logs (×3): `sleep 600; echo waited` consumed the entire 600s exec budget and timed out — a sleep that long can never return within the cap, so the timeout was certain at the moment of writing the command. Sleeping to poll is legitimate; sleeping past the budget is arithmetic failure. Block only sleeps that cannot fit: the structural layer tokenizes argv, so it can parse `sleep <secs>` and compare against the configured exec timeout (minus headroom for whatever follows the sleep). Guidance on the block: state the budget, suggest shorter poll intervals, and name github_ci_status for the CI case specifically.

kitaebot:
Maximum iterations reached without completion. State at exit:
State report for successor — issue amackillop/kitaebot#136 (deny-list guidance + sleep rule)

## What was done

**Branch:** `kitaebot_issue-136_deny-guidance` (checked out at `projects/amackillop/kitaebot`, branched from origin/HEAD 4644f0d). **No commits yet** — all work is uncommitted in the working tree:
- `src/tools/exec.rs` — modified (~200 lines diff)
- `specs/03-tools.md` — modified (structural-layer sleep paragraph + guidance-policy paragraph)
- `src/tools/dbg_t.rs` — **scratch debug module, must be deleted** before commit
- `src/tools/mod.rs` — has a stray `#[path = "dbg_t.rs"] mod dbg_t;` line appended at the end — **must be removed**

## Completed and verified

[...]

PR #140, opened 2026-09-02
Decide exec deny command position with real quoting

### LINKED ISSUE #135: Exec deny regexes false-positive on denied names inside quoted arguments
Repro from the 2026-09-01 logs (turn working #121): `grep -n "index_over_cap\|compaction\|truncate\|marker..." src/memory/mod.rs` was blocked. The quoted grep pattern contains the literal `\|truncate`, and the deny rule for the `truncate` binary anchors to command position — start or after a separator — which matches the `|` inside the quoted argument. The regex layer cannot see quoting, so any grep/rg alternation naming a denied binary (truncate, shred, dd, mount, ...) is blocked. The bot was denied grepping its own source for the word truncate while working the issue that required it.

This is the same failure class the bare-name anchoring commit already fixed for prose in quoted arguments; alternation metacharacters reopened it. The structural shlex layer tokenizes with real quoting rules, so command-position semantics belong there: candidates are moving the binary-name rules to the structural layer entirely, or having the regex layer pre-strip quoted spans before matching (lossy but cheap). Whichever way, the fix needs a regression test with a quoted alternation naming every structurally-denied binary.

amackillop:
Closes #135.

The exec deny regexes anchor bare-name rules to command position with
`(^|[|;&\n])`, which matches the `|` inside a quoted grep alternation.
The bot was blocked grepping its own source for the word truncate. The
regex layer cannot see quoting, so this is unfixable where it lives:
command position has to be decided after real quote handling.

Two commits, in dependency order:

1. **Split structural deny segments with quote awareness.** The
   structural layer used to split on operator *tokens*, which only
   works when the operator is whitespace-separated (`x|truncate` folds
   into one token) and wrongly treats a quoted `"|"` argument as an
   operator once shlex strips quotes. It now splits the raw string on
   unquoted `|`, `;`, `&`, and newline, tracking quote and backslash
   state, then shlex-parses each segment. This has to land first:
   without it, deleting the regex rules would open unspaced-pipe and
   second-line bypasses the regexes currently mask.

2. **Move the bare-name deny rules to the structural layer.** The
   eleven command-position regexes (shred, wipe, truncate, mount,
   umount, shutdown, reboot, poweroff, halt, su, at) become binary-only
   structural rules. This preserves d3117b2's intent (commands, never
   words) with real quoting rules, and closes the env-prefix
   (`FOO=bar truncate`) and path-prefix (`/usr/bin/shred`) bypasses the
   anchored regexes had. Regression test covers the verbatim #135 repro
   plus one alternation naming every moved binary, in both BRE and ERE
   form. Spec 03 updated; it also stops crediting the whole Nix deny
   list to the structural layer (only `nix profile` lives there).

Deliberate tightenings: bare `su` and bare `at` are now blocked (the
old regexes required a trailing space). Known non-goals: the anchored
nix remote-flake regexes share the quoted-false-positive class but need
argument matching the structural layer cannot express; command
substitution (`$(...)`) remains invisible to both layers, as before.

### DIFF
--- a/src/tools/exec.rs
+++ b/src/tools/exec.rs
@@ -75,22 +75,6 @@ const DENY_RULES: &[DenyRule] = &[
         pattern: r"\bfind\b.*-exec\s+rm\b",
         guidance: BLOCKED,
     },
-    // Bare-name commands are anchored to command position (start or
-    // after a separator): \b alone matches the same names as English
-    // prose inside quoted arguments — a self-analysis turn was once
-    // halted for echoing "not at workspace root".
-    DenyRule {
-        pattern: r"(^|[|;&\n])\s*shred\b",
-        guidance: BLOCKED,
-    },
-    DenyRule {
-        pattern: r"(^|[|;&\n])\s*wipe\b",
-        guidance: BLOCKED,
-    },
-    DenyRule {
-        pattern: r"(^|[|;&\n])\s*truncate\b",
-        guidance: BLOCKED,
-    },
     // Disk / filesystem
     DenyRule {
         pattern: r"\bmkfs\b",
@@ -109,34 +93,10 @@ const DENY_RULES: &[DenyRule] = &[
         guidance: BLOCKED,
     },
     DenyRule {
-        pattern: r"(^|[|;&\n])\s*mount\b",
-        guidance: BLOCKED,
-    },
-    DenyRule {
-        pattern: r"(^|[|;&\n])\s*umount\b",
-        guidance: BLOCKED,
-    },
-    DenyRule {
         pattern: r"(^|[^0-9])>\s*/dev/",
         guidance: BLOCKED,
     },
     // System power
-    DenyRule {
-        pattern: r"(^|[|;&\n])\s*shutdown\b",
-        guidance: BLOCKED,
-    },
-    DenyRule {
-        pattern: r"(^|[|;&\n])\s*reboot\b",
-        guidance: BLOCKED,
-    },
-    DenyRule {
-        pattern: r"(^|[|;&\n])\s*poweroff\b",
-        guidance: BLOCKED,
-    },
-    DenyRule {
-        pattern: r"(^|[|;&\n])\s*halt\b",
-        guidance: BLOCKED,
-    },
     DenyRule {
         pattern: r"\binit\s+[0-6]\b",
         guidance: BLOCKED,
@@ -148,10 +108,6 @@ const DENY_RULES: &[DenyRule] = &[
     // Privilege escalation
     DenyRule {
         pattern: r"\bsudo\b",
-        guidance: BLOCKED,
-    },
-    DenyRule {
-        pattern: r"(^|[|;&\n])\s*su\s",
         guidance: BLOCKED,
     },
     DenyRule {
@@ -330,10 +286,6 @@ const DENY_RULES: &[DenyRule] = &[
     // Cron / persistence
     DenyRule {
         pattern: r"\bcrontab\b",
-        guidance: BLOCKED,
-    },
-    DenyRule {
-        pattern: r"(^|[|;&\n])\s*at\s",
         guidance: BLOCKED,
     },
     // Git operations that must go through their dedicated tools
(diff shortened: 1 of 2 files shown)


src/tools/exec.rs
```diff
@@ -875,38 +885,58 @@ fn is_env_assignment(token: &str) -> bool {
...
+            escaped = false;
+        } else {
+            match (&state, c) {
+                (QuoteState::Double, '"') | (QuoteState::Single, '\'') => {
+                    state = QuoteState::None;
+                }
+                (QuoteState::Double | QuoteState::None, '\\') => escaped = true,
+                (QuoteState::None, '"') => state = QuoteState::Double,
+                (QuoteState::None, '\'') => state = QuoteState::Single,
+                (QuoteState::None, '|' | ';' | '&' | '\n') => {
```
kitaebot, line 916: should-fix: a single `&` (background) is treated as a segment separator, so `true& truncate`-style inputs are rejected with "unparseable shell syntax" only via accidental shlex behavior rather than deterministic splitter logic. Pre-existing ambiguity versus `&&`; splitting per character position would make the `&`/`&&` distinction explicit instead of an artifact of where shlex gives up.

src/tools/exec.rs
```diff
@@ -1733,4 +1735,65 @@ mod tests {
...
+    fn split_unquoted_separators_splits_on_bare_operators() {
+        assert_eq!(
+            split_unquoted_separators("a | b; c && d"),
+            vec!["a ", " b", " c ", "", " d"]
+        );
+        assert_eq!(split_unquoted_separators("a\nb"), vec!["a", "b"]);
+        assert_eq!(split_unquoted_separators("a|b"), vec!["a", "b"]);
+    }
+
+    #[test]
```
kitaebot, line 1749: nit: the escaped-quote / backslash-backslash / escape-reset path in the splitter has no direct test. Coverage pins quoted alternations and operators-stay-arguments, but the escape semantics are what make those safe — one test locking them down would guard against drift in the delimiter set.

kitaebot (approved):
Verdict: **correct** — the raw-string quote-aware splitter is the right design (the decision has to happen before shlex; once quotes are stripped a literal `|` argument is indistinguishable from an operator), and moving the eleven bare-name rules to the structural layer preserves d3117b2's intent while closing the env-prefix and path-prefix bypasses the anchored regexes had.

Reviewer note: the reviewer's one blocking finding — that this PR deletes the #136 oversized-sleep guard against a stale base — was checked against git and does not hold: the merge base is current master HEAD (4644f0d), and `oversized_sleep`/`SLEEP_HEADROOM_SECS`/`sleep_secs` appear in neither master nor this branch's `exec.rs`. Nothing was deleted. The remaining findings are non-blocking and inline.

Test coverage on the moved rules is better than the repo norm — verbatim #135 repro, all eleven binaries in one alternation in both BRE and ERE, unspaced-pipe, second-line, and quoted-operator-stays-argument cases.

src/tools/exec.rs
amackillop: Checked, and the premise does not hold. The splitter already works per character position: a single & ends a segment, and && is two separators with an empty segment between (skipped). `true& truncate -s 0 f` never reaches the unparseable fallback — the second segment is `truncate -s 0 f`, shlex parses it fine, and the truncate rule fires. 74c9c53 pins this with assert_eq on the exact guidance returned (`Some(BLOCKED)`, not "unparseable shell syntax"), so if the mechanism ever degrades to the fallback the test names it.

src/tools/exec.rs
amackillop: Fair. 74c9c53 adds split_unquoted_separators_escape_semantics covering the three rules the guarantees ride on: escaped quote does not close the string, double backslash is a complete escape so a following separator still splits, and backslash inside single quotes is literal.

src/tools/exec.rs
kitaebot: Conceded — the premise was wrong. The splitter walks per character: unquoted `&` ends the segment and `&&` is two separators with an empty one between, so `true& truncate -s 0 f` splits deterministically before shlex runs and the truncate rule fires with `Some(BLOCKED)`. Verified against the code; the assert_eq on the exact guidance means a regression to the unparseable fallback fails the test by name. Good pin.

src/tools/exec.rs
kitaebot: Confirmed — `split_unquoted_separators_escape_semantics` covers all three rules, each against a distinct match arm: escaped quote doesn't close the Double state, double backslash completes the escape so the following `;` still splits, and backslash inside single quotes falls to the literal arm. Right outcomes asserted in each.

Issue #139, opened 2026-09-02
Summarizer and raw-chat provider calls never reach the usage ledger


amackillop:
## Problem

The summarizer closure and the cache-prefix-riding raw-chat closure (`context/mod.rs` — `make_summarize_fn`, `make_raw_chat_fn`) discard `ChatOutcome.usage` even on success: they return `Result<String, ProviderError>`, so the tokens and cost of every compaction and distillation call are invisible to the ledger. #128 fixes failed-attempt usage at the provider boundary, but these call sites drop *successful* usage too — a compaction-heavy turn under-reports by its full summarization cost.

## Proposal

Thread usage from these closures back to the turn that triggered them (compaction fires inside `turn_loop`/between turns, so a `TurnUsage` accumulator is usually in reach), or account them as their own ledger rows under a synthetic task. Needs a design decision on attribution: bill the owning turn vs. a standalone `summarizer` row.

Split from #128 (failed-attempt usage), which lands the `ChatOutcome.failed`/`ChatError` plumbing these sites can reuse.

Issue #120, opened 2026-08-28
Empty provider response kills the subagent and discards its context; parent respawns from zero


amackillop:
## Symptom

Three explore subagents in turn 6 (Duty, `CumuloGlobal/lightning-node`, 2026-08-28) died the same way, minutes of accumulated context each, all discarded:

```
18:58:49 ERROR subagent{explore}: Tool execution failed: Blocked: fetch https://api.github.com/.../dependabot/alerts/750 (HTTP 403) tool=web_fetch
18:58:55 WARN  subagent{explore}: Provider returned empty content with no tool calls finish_reason="stop" reasoning_len=0
18:58:55 ERROR Tool execution failed: sub-agent failed: Provider error: Provider returned an empty response tool=task
```

Same sequence at 18:31:35 and 19:01:14. Each time the parent's answer was to spawn a fresh explore with the same brief, which re-derived everything and then hit the identical wall. Three wasted runs at 5–10 minutes apiece in a single turn.

## Problem

A 200 response with empty content and `finish_reason="stop"` is a transient provider fault (deepseek-v4-pro via OpenRouter here), but it's currently terminal for the subagent: the whole conversation is thrown away and the failure surfaces to the parent as a failed `task` tool call.

## Fix

Retry the *completion*, not the subagent. Treat empty-content/no-tool-calls as retryable in the provider retry unit (the same layer that already reparses inside the retry loop) — one more request against a context that's already built beats respawning an agent that rebuilds it from nothing. Bound it (2–3 attempts) so a provider that's genuinely wedged still fails through.

Secondary observation from the same trace: the blocked web_fetch that precedes each death suggests the explore tool scope lacks the `github_*` tools the block message recommends. Worth checking, but the retry gap is the bug — any tool error followed by a provider hiccup produces the same total loss.

amackillop:
Chesterton's fence check: commit 83dbf35 deliberately keeps `EmptyResponse` out of `is_transient` — "Truncated and EmptyResponse stay out of is_transient and keep failing on the first attempt. There is a test pinning that, because the blast radius of getting it wrong is retrying a starved reasoning model three times at full prompt cost."

[...]

PR #141, opened 2026-09-02
Keep billed usage from failed provider attempts

### LINKED ISSUE #128: Failed provider attempts drop their billed usage from the ledger

amackillop:
## Problem

A failed provider attempt returned `Err(ProviderError)` with no usage attached, so tokens billed for the draw never reached the ledger (#128). `attempt()` built the `CallUsage` before parsing, then discarded it with the parse failure. The 2026-09-01 incident: an `EmptyResponse` after 3317 billed reasoning tokens, invisible in `/usage`. #127 making empties retryable widened the hole, and per-turn cost comparisons were flattered by dropping the cost of exactly the failures cheap models produce more of.

## Change

- Parse-shaped failures (`EmptyResponse`, `Truncated`, `MalformedToolCall`, empty-choices `InvalidResponse`) carry their usage out of `attempt()`; transport failures billed nothing and carry nothing.
- `chat()` accumulates failed-draw usage across redraws: `ChatOutcome` grows a `failed` vector, exhaustion returns `ChatError { error, failed }`. Accumulation lives outside the retry unit, so the combinator and policy stay on plain `ProviderError` and the #120 parse-inside-retry fence is untouched.
- The turn loop folds `failed` into `TurnUsage` on both arms, so failed draws count as calls in the ledger. Only the successful draw's `prompt_tokens` feeds context-size observation.
- Wrapper over usage fields on `ProviderError` variants: keeps the error enum a pure failure-mode ADT, and the final error could only carry its own attempt's usage anyway.
- Specs 02 and 27 updated.

Landed as one commit rather than the planned two: the trait signature change forces the agent-loop edits regardless, and staging `ChatOutcome.failed` without its consumer trips `dead_code`.

The summarizer/raw-chat closures dropping usage even on success is pre-existing, split to #139.

Closes #128

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,13 +4,14 @@ Autonomous programming agent in Rust. Runs in a NixOS VM with Landlock sandboxin
 
 ## Overview
 
-Kitaebot is a long-running daemon that accepts messages via Telegram, Unix socket, GitHub PR comments, GitHub issues, or Linear issues, routes them through an LLM agent loop with tool use, and persists conversation state through a pluggable context engine. A duty scheduler runs recurring work on its own schedule.
+Kitaebot is a long-running daemon that accepts messages via Telegram, Unix socket, GitHub PR comments, GitHub issues, or Linear issues, routes them through an LLM agent loop with tool use, and persists conversation state through a pluggable context engine. A duty scheduler runs recurring work on its own schedule. The agent keeps durable cross-session knowledge in a memory subsystem, reviews its own work through gated reviewer sub-agents, and can consume external MCP tool servers.
 
 Two binaries:
 
 | Binary | Purpose | Lifecycle |
 |--------|---------|-----------|
 | `kitaebot run` | Daemon (Telegram + socket + duties + GitHub PRs/issues + Linear) | systemd service |
+| `kitaebot backup <dir>` | Stage durable workspace state into `<dir>` for archiving | On-demand |
 | `kchat <socket>` | Socket client REPL | On-demand |
 
 ## Architecture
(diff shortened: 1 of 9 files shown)


kitaebot (approved):
Verdict: **correct** — the failed-draw usage accounting holds end-to-end and I could not find a path that still drops billed tokens.

The core claims check out against the code:

- `attempt()` binds `CallUsage` before `parse_response`, so parse-shaped failures genuinely carry it out; transport failures carry `None`, matching the spec.
- `chat()` accumulates each failed attempt exactly once into the `failed` vector, outside the retry unit — the #120 parse-inside-retry fence is untouched and the policy still sees plain `ProviderError`.
- Both turn-loop arms fold `failed` into `TurnUsage`, and the meter is built from `stats.usage` after `turn_loop` returns regardless of result, so error/cancel paths still reach the ledger. Only the successful draw's `prompt_tokens` feeds context-size observation — no double-count, no inflation.
- The discarded-usage wrapper (`run_turn`) is `#[cfg(test)]`-only; all production paths go through `run_turn_metered`.
- Specs 02 and 27 match the code, and the new tests vary their parameters away from defaults rather than asserting defaults.

The wrapper-over-fields choice over usage fields on the `ProviderError` variants is the right call: the error stays a pure failure-mode ADT and the final error could only carry its own attempt's usage anyway.

The README/`configuration.nix`/model-override commits on the branch are docs and config, unrelated to this fix — nothing objectionable.

PR #142, opened 2026-09-02
Recover unparseable reviewer findings with one corrective retry, and make drops visible

### LINKED ISSUE #113: Unparseable reviewer findings block is silently dropped; parent cites nonexistent ledger ids
## Symptom (from the 2026-08-27 duty journal)

```
03:06:55 WARN  reviewer response has no parseable findings block   (kitaebot::agent::task)
03:57:51 WARN  reviewer response has no parseable findings block   (kitaebot::agent::task)
03:58:01 ERROR Tool execution failed: Invalid arguments: no finding with id 121  (review_disposition)
03:58:05 ERROR Tool execution failed: Invalid arguments: no finding with id 122  (review_disposition)
```

Both WARNs are reviewer sub-agent calls whose responses ended in a fenced `findings` block in the plain `verdict: correct` / `findings: - id: ...` shape rather than the JSON object the parser expects. The findings were real (two nits, later fixed by the author), but they were never recorded in the ledger. The parent agent — which cannot see the journal — then cited "ledger findings 121, 122" in a posted GitHub PR review (ids that did not exist) and only discovered the drop when `review_disposition` rejected them.

## Mechanism

- `src/review.rs:74-79` — `parse_findings_block` accepts only JSON inside the ` ```findings ` fence; any other shape returns `None`.
- `src/agent/task.rs:379-407` — `record_review` treats a parse failure as telemetry-only: WARN to the journal, return `Vec::new()`. The doc comment states the contract: "Telemetry only: every failure is a warning, never a review failure."
- `src/agent/task.rs:360-368` — the `[ledger: finding ids ...]` trailer is appended to the task result **only when ids were recorded**. On parse failure the returned text carries no signal that recording was attempted and failed.

So the task result is indistinguishable between "findings recorded" and "findings dropped", while the ledger ids are load-bearing downstream: `review_disposition` requires them, and the review workflow cites them in PR reviews. The failure isolation is correct for the *review* (the verdict prose still reaches the parent), but it silently breaks the *disposition loop*.

The format itself is pinned only by prompt: `src/prompts/reviewer.md:66-80` mandates the JSON shape, but a parent dispatch prompt that asks for a different shape can override it — which is exactly what happened here. LLM output format is never guaranteed regardless.

## Fix directions (fork needing a call)

1. **Surface the failure to the parent**: append a trailer like `[ledger: findings block unparseable; nothing recorded]` when `record_review` drops a block, mirroring the success trailer. Minimal, keeps the parser strict.
2. **Tolerant parser**: accept the plain `verdict:` / `findings:` shape in addition to JSON. Removes the failure class but adds a second format to maintain.
3. Both.

Option 1 alone fixes the misleading behavior (the parent knows to re-derive or skip the ids); option 2 additionally keeps the ledger complete. Not fixed here per triage policy.

kitaebot:
Closes #113

## Problem

When a reviewer response's fenced `findings` block was absent or malformed, `record_review` dropped it silently: a journal WARN, no ledger rows, and a task result byte-identical to a successful recording. The parent — which cannot see the journal — cited ledger ids that were never inserted and discovered the drop only when `review_disposition` rejected them. That was the 2026-08-27 incident: two real findings lost, two phantom ids cited in a posted PR review.

## Fix

- **One corrective retry.** On an unparseable block with the ledger enabled, one corrective turn runs on the reviewer's *own* ephemeral session (the reviewer sees its prior text and re-emits the block), then the text is re-parsed. No fresh re-review, no config knob — the same bounded-retry doctrine as the distill compaction retry.
- **Failure trailers.** A residual parse failure appends `[ledger: findings block unparseable; nothing recorded]`; a failed ledger write appends `[ledger: recording failed; nothing recorded]`. "Recorded" and "dropped" result text is never identical again, and a failure trailer tells the parent no ids exist to disposition. `review-gates.md` and `review-protocol.md` state this.
- **Parser unchanged** — strict. Accepting a second prose format would license every future shape drift as "supported"; the retry handles the failure class instead.
- **Retry spend folds into the gate invocation's single usage row** via new `TurnUsage::add_turn` (same fold the distill retry uses).
- **Bill-before-propagate preserved**: `record_turn` for a failed sub-agent moved into the error arm, so an errored sub-agent still bills before the error reaches the parent.

Spec 23's failure-mode rows and identity paragraph updated to match.

## Tests

Six new tests in `agent::task`: retry-then-record (2 provider calls, ids trailer on the original prose), residual unparseable after retry (trailer, zero ledger rows), provider-error retry (review still delivered), disabled ledger (no retry, no trailer), retry prompt content, and ledger-write failure (trailer, no ids, no retry — via a dropped `reviews` table behind the ledger's back). The malformed fixture is the incident's real prose shape.

`just check` passes. Plan was posted to the ticket and approved; commit and series gates both run, findings fixed and dispositioned.

### DIFF
--- a/src/agent/task.rs
+++ b/src/agent/task.rs
@@ -90,13 +90,8 @@ const FINDINGS_RETRY_PROMPT: &str = "Your previous response had no parseable ```
 block (exactly one fenced block holding a single JSON object). Re-emit your review's \
 verdict and findings as exactly one such block, unchanged in substance, nothing else.";
 
-/// The corrective turn runs with no tool definitions and one
-/// iteration: it can only answer with text, at most one provider call
-/// plus one `FinalAnswer` squeeze. Session histories that hold tool
-/// rounds still carry their `tool_calls` and providers like GLM reject
-/// such a history when the tools are absent from the request; the
-/// failure is soft — a warning and the unparseable trailer, the
-/// parent's text survives.
+/// Toolless and single-iteration: the corrective turn can only
+/// answer with text (see the commit for the GLM tool-history caveat).
 const FINDINGS_RETRY_MAX_ITERATIONS: usize = 1;
 
 /// A sub-agent type: system prompt plus prebuilt tool set.
--- a/src/agent/task.rs
+++ b/src/agent/task.rs
@@ -1230,6 +1230,60 @@ mod tests {
         assert!(!result.contains("[ledger:"), "{result}");
     }
 
+    /// Exercise the ledger-write failure path: the reviews table is
+    /// dropped behind the ledger's back, so `record_review` fails at
+    /// the statement — the same `Err` any schema/IO failure surfaces.
+    #[tokio::test]
+    async fn failed_ledger_write_carries_its_trailer() {
+        let dir = tempfile::tempdir().unwrap();
+        let ledger = Arc::new(crate::review::ReviewLedger::new(
+            &crate::state_db::StateDb::open(&dir.path().join("kitaebot.db")).unwrap(),
+        ));
+        let review = "Bad.\n```findings\n\
+             {\"verdict\": \"incorrect\", \"confidence\": 0.9, \
+             \"explanation\": \"x\", \
+             \"findings\": [{\"category\": \"c\", \"note\": \"n\"}]}\n```";
+        let provider = Arc::new(MockProvider::new(vec![Ok(Response::Text(
+            review.to_string(),
+        ))]));
+        {
+            let conn = ledger.connection_for_test();
+            conn.execute("DROP TABLE reviews", []).unwrap();
+        }
+        let tool = TaskTool::new(
+            same_provider(&provider),
+            noop_summarize(),
+            AgentTypes {
+                explore: agent_type(Tools::default()),
+                worker: agent_type(Tools::default()),
+                reviewer: agent_type(Tools::default()),
+            },
+            5,
+            None,
+            Some(ledger.clone()),
+        );
+        let result = tool
+            .execute(
+                serde_json::json!({
+                    "prompt": "review this",
+                    "agent_type": "reviewer",
+                    "review": {"repo": "o/r", "gate": "commit", "git_ref": "abc"}
+                }),
+                ToolCtx::default(),
+            )
+            .await
+            .unwrap();
+        assert_eq!(provider.call_count(), 1, "no retry on a write failure");
+        assert!(
+            result.contains("[ledger: recording failed; nothing recorded]"),
+            "{result}"
+        );
+        assert!(
+            !result.contains("[ledger: finding ids"),
+            "a failed write must not cite ids: {result}"
+        );
+    }
+
     #[tokio::test]
     async fn clean_review_gets_no_id_trailer() {
         let dir = tempfile::tempdir().unwrap();
--- a/src/agent/task.rs
+++ b/src/agent/task.rs
@@ -27,7 +27,7 @@ use crate::tools::mcp::McpTools;
 use crate::tools::{Tool, ToolCtx, Tools, string_or_value};
 use crate::usage::{self, TurnRecord, UsageLedger};
 
-use super::{BudgetPolicy, ReplyPolicy, run_turn_metered};
+use super::{BudgetPolicy, ReplyPolicy, TurnMeter, run_turn_metered};
 
 /// Allowlist for the `explore` type: read-only research.
 ///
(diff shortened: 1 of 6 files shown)


amackillop (commented):
Mergeable. Mechanism is right, tests cover the real incident shape, bill-before-propagate is intact. Three non-blocking notes inline; the first is the one worth acting on.

src/agent/task.rs
```diff
@@ -352,37 +406,119 @@ impl<P: Provider> Tool for TaskTool<P> {
...
+                    ctx,
+                    ..
+                } = retry;
+                let (result, meter) = run_turn_metered(
+                    engine,
+                    &self.summarize,
+                    system_prompt,
+                    FINDINGS_RETRY_PROMPT,
+                    provider,
+                    tools,
```
amackillop: This is not "one corrective turn" in any bounded sense. The retry gets the full reviewer tool set and the full `max_iterations`. The `FinalAnswer` degraded-answer case is exactly the one that produces a blockless response, so a reviewer that already burned its budget gets a second full budget to go researching again.

Pass `&Tools::default()` here. No tool definitions means the provider can only answer with text, so the corrective turn becomes literally one call, which is what the spec row promises.

Verify first: the session history holds prior `tool_use` blocks, and some providers reject a request whose history references tools absent from the current definitions. Anthropic accepts it. Check GLM.

src/agent/task.rs
```diff
@@ -352,37 +406,119 @@ impl<P: Provider> Tool for TaskTool<P> {
...
+                        let retry_text = output.into_text();
+                        (record_parse(ledger, retry.meta, &retry_text), Some(meter))
+                    }
+                    Err(e) => {
+                        warn!("findings retry turn failed: {e}");
+                        (ReviewRecording::Unparseable, Some(meter))
+                    }
+                }
+            }
+            ReviewRecording::Disabled => (ReviewRecording::Disabled, None),
```
amackillop: Dead arm. `record_parse` takes `&ReviewLedger`, so it can never return `Disabled`, yet this match has to name it. Two types make the impossible state unrepresentable:

```rust
enum ParseOutcome { Recorded(Vec<i64>), Unparseable, WriteFailed }
enum ReviewRecording { Disabled, Parsed(ParseOutcome) }
```

Cosmetic.

src/agent/task.rs
```diff
@@ -338,8 +345,55 @@ impl<P: Provider> Tool for TaskTool<P> {
...
+                            meta: args.review.as_ref(),
+                            provider,
+                            system_prompt: &agent.system_prompt,
+                            tools: &agent.tools,
+                            max_iterations: self.max_iterations,
+                            ctx: &child_ctx,
+                        },
+                        &text,
+                    )
+                    .instrument(tracing::info_span!("subagent", agent = "reviewer-retry"))
```
amackillop: This span wraps the whole of `record_review_with_retry`, including the first parse, so the "no parseable findings block; retrying once" warning logs under `reviewer-retry` before any retry exists. Move the `.instrument` onto the inner `run_turn_metered` call.

src/agent/task.rs
kitaebot: Fixed in 003d7df: the corrective turn now runs with `&Tools::default()` and `FINDINGS_RETRY_MAX_ITERATIONS = 1`, so it is structurally one turn — its only possible output is text (at most one call plus the FinalAnswer squeeze). The `ReviewRetry` struct lost its `tools`/`max_iterations` fields.

GLM check: GLM (per Zhipu's OpenAI-compatible API behavior) rejects a history holding assistant `tool_calls` when the tools are absent from the request — unlike Anthropic. That matters for the subcase where the reviewer ran tool rounds and its final text was still unparseable: the retry session then carries `tool_calls` and the toolless retry request would 400. It fails soft — warn + unparseable trailer, parent's text survives — and the new test pins the toolless wire shape via a per-call tools capture added to `MockProvider` (`request_tools(index)`).

src/agent/task.rs
kitaebot: Fixed in 003d7df: split into `ParseOutcome { Recorded, Unparseable, WriteFailed }` and `ReviewRecording { Disabled, Parsed(ParseOutcome) }` — `record_parse` returns `ParseOutcome` only, so the dead `Disabled` arm is unrepresentable.

src/agent/task.rs
kitaebot: Fixed in 003d7df: the outer `.instrument` is gone; the corrective `run_turn_metered` call carries its own `findings-retry` span, so the first warning now logs under the parent reviewer span.

amackillop:
Rebase on master with git_rebase and resolve the conflict in src/agent/mod.rs.
Master already has TurnUsage::add_turn (195fd19); drop this branch's copy and
keep master's. Run `just check` before `continue`.

amackillop (approved):

PR #143, opened 2026-09-02
Drain the in-flight turn on SIGTERM

### LINKED ISSUE #133: SIGTERM hard-kills the in-flight turn; drain it instead
## Problem

Every daemon stop is a hard kill of in-flight work: shutdown is a `tokio::select!` (daemon.rs `run_with_shutdown`) that drops the actor task, the running turn, its sub-agents, and any mid-dispatch channel tick the instant SIGTERM lands. Operator-initiated restarts — deploys, state surgery, disk recovery — are nearly all restarts, and each one has cost real work: a 20-iteration review turn killed during the #114 anchor repair, whatever was running when the disk-full VM went down, and every rebuild to come.

## Design

The actor pattern already contains the graceful exit, unused: `Agent::run` returns naturally when all `AgentHandle` senders drop, after finishing its current envelope. The drain is rearrangement plus one flag:

1. `AgentHandle::spawn` surfaces the actor's `JoinHandle` instead of discarding it.
2. On the shutdown signal, `run_with_shutdown` exits the `select!` (dropping the channel loops and socket listener — they stop producing, and a dropped poll tick never saved its cursor, so its events re-dispatch on the next boot), drops the handle, and awaits the actor's join instead of returning.
3. A shared draining flag, set at shutdown: the actor finishes the in-flight turn normally but replies an error to envelopes already queued in the mailbox without running them. Without this, up to one queued turn per channel would run to completion with its reply channel dead — work done, comment never posted, then re-dispatched next boot into a session that already contains it. Channels are dropped so the error goes nowhere; the point is skipping cheaply.
4. `TimeoutStopSec` on the service (vm/configuration.nix) raised to cover one long turn (~30 min; the flash default has produced 12-minute turns), so systemd does not SIGKILL mid-drain; a genuinely hung turn still dies at the bound. Every stop/rebuild flow inherits the drain for free.

Sub-agents need nothing: they are awaited inside the turn and finish with it.

## Out of scope

Genuine crashes (OOM, qemu death) are not drainable; their recovery is channel re-dispatch consistency, which is complete for issues/Linear (cursor saved post-dispatch) but not for tracked-PR reviews (reviewed SHA recorded at dispatch, so a killed review never re-runs) — separate issue.

Note for the plan: the drain flag and the daemon wiring must land as one commit; a `begin_drain` nobody calls fails the dead-code lint (same lesson as TurnRole::Planner in #122).

amackillop:
Closes #133.

Every daemon stop hard-killed the running turn: `run_with_shutdown` returned out of its `select!` on SIGTERM and took the actor task with it. The actor pattern already exits gracefully once every sender is gone, after finishing the envelope in hand. This PR uses that.

## Commits

1. **Drain the in-flight turn on SIGTERM instead of killing it.** `AgentHandle::spawn` returns the actor's `JoinHandle`; the daemon takes the handle by value, and on shutdown drops the channel loops, calls `begin_drain()`, drops the handle, and awaits the task. A draining actor refuses envelopes already queued in the mailbox with `Err("Agent draining")` (their reply channels are dead, and the channel re-dispatches the event next boot). Sub-agents are awaited inside the turn and finish with it. `MockProvider::gated` holds a turn in flight for the test.
2. **Give the systemd unit room to drain the turn.** `TimeoutStopSec=30min` so systemd does not SIGKILL mid-drain. `KillMode=mixed` so SIGTERM reaches the daemon alone; the default control-group mode would kill the exec tool's children and the MCP servers under the turn being drained.

Specs 01 and 09 and the README updated.

## Verify after deploy

1. Start a long turn (`just ask` or a Telegram message).
2. `systemctl restart kitaebot` in the VM.
3. Journal: `Shutdown signal received.` then `Draining: waiting for the in-flight turn to finish.` then the turn summary, then `Drain complete, exiting.` The reply lands on the channel.

## Note

The first `just check` run hit an unrelated flake: `clients::telegram::tests::network_error_redacts_token_url` bind-and-drops an ephemeral port and asserts "Connection refused", but got "Connection reset by peer" when a parallel test reused the port. Pre-existing; worth its own issue.

### DIFF
--- a/specs/01-agent-loop.md
+++ b/specs/01-agent-loop.md
@@ -281,6 +281,18 @@ hint for display formatting.
 
 If the actor has shut down, `send_message` returns a synthetic error string.
 
+**Shutdown drains.** `AgentHandle::spawn` also returns the actor task's
+`JoinHandle`. On SIGINT/SIGTERM the daemon stops the channel loops, calls
+`begin_drain()`, drops its handle, and awaits the task. The actor finishes the
+turn it is on (sub-agents are awaited inside the turn, so they finish with it),
+answers every envelope already queued with `Err("Agent draining")` instead of
+running it, and exits when the last sender is gone. Refusing the queue matters:
+its reply channels are dead, so running it would do the work, post nothing, and
+re-dispatch the same event on the next boot into a session that already
+contains it. Channel events lost this way re-dispatch because a dropped poll
+tick never saved its cursor. The systemd unit's `TimeoutStopSec` bounds the
+drain (spec 09); a hung turn still dies there.
+
 ### ChannelSource
 
 Messages are tagged with their origin before entering the session:
--- a/specs/01-agent-loop.md
+++ b/specs/01-agent-loop.md
@@ -284,7 +284,7 @@ If the actor has shut down, `send_message` returns a synthetic error string.
 **Shutdown drains.** `AgentHandle::spawn` also returns the actor task's
 `JoinHandle`. On SIGINT/SIGTERM the daemon stops the channel loops, calls
 `begin_drain()`, drops its handle, and awaits the task. The actor finishes the
-turn it is on (sub-agents are awaited inside the turn, so they finish with it),
+queued input envelopes with `Err("Agent draining")`
 answers every envelope already queued with `Err("Agent draining")` instead of
 running it, and exits when the last sender is gone. Refusing the queue matters:
 its reply channels are dead, so running it would do the work, post nothing, and
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ Channels (Telegram, Unix socket, GitHub PRs, GitHub issues, Linear, Duties)
            messages tagged by source)
 ```
 
-The agent is an actor (Ryhl pattern) — a spawned tokio task that processes one envelope at a time. Channels hold cloneable `AgentHandle`s and send messages via `send_message()`, awaiting a reply over a oneshot channel. This eliminates session locking: the actor owns the session and processes requests sequentially.
+The agent is an actor (Ryhl pattern) — a spawned tokio task that processes one envelope at a time. Channels hold cloneable `AgentHandle`s and send messages via `send_message()`, awaiting a reply over a oneshot channel. This eliminates session locking: the actor owns the session and processes requests sequentially. SIGTERM drains instead of killing: the channel loops stop, the in-flight turn runs to completion, queued envelopes are refused, then the daemon exits.
 
 The agent loop calls the LLM, dispatches tool calls in parallel, checks outputs for leaked secrets, and repeats until the model produces a final response or hits `max_iterations`.
 
--- a/specs/09-vm.md
+++ b/specs/09-vm.md
@@ -53,6 +53,10 @@ The daemon runs as a systemd service (`Type=simple`):
 - **WorkingDirectory**: `/var/lib/kitaebot`
 - **RuntimeDirectory**: `kitaebot` (creates `/run/kitaebot/` for the socket)
 - **Restart**: `on-failure`, 10s delay
+- **Stop**: `KillMode=mixed`, `TimeoutStopSec=30min`. SIGTERM reaches the
+  daemon alone, which drains the in-flight turn (spec 01); tool children and
+  MCP servers keep running under it and are SIGKILLed with anything left once
+  it exits. A turn still running at the bound is killed.
 - **TasksMax/LimitNPROC**: 4096
 - **Environment**: `KITAEBOT_WORKSPACE`, `RUST_LOG`, `PATH` (restricted to
   the runtime dependencies plus `tools` packages' bin dirs),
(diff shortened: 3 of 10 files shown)


specs/01-agent-loop.md
```diff
@@ -281,6 +281,18 @@ hint for display formatting.
 
 If the actor has shut down, `send_message` returns a synthetic error string.
 
+**Shutdown drains.** `AgentHandle::spawn` also returns the actor task's
+`JoinHandle`. On SIGINT/SIGTERM the daemon stops the channel loops, calls
+`begin_drain()`, drops its handle, and awaits the task. The actor finishes the
+turn it is on (sub-agents are awaited inside the turn, so they finish with it),
```
kitaebot: This overclaims: greeting envelopes queued at drain time are answered with a greeting, not `Err("Agent draining")` — the commit message itself says so ("Greetings still get answered"). Per the specs-are-contracts convention, scope this to queued *input* envelopes:

```suggestion
queued input envelopes with `Err("Agent draining")`
```

(adjust surrounding words as needed — the claim is about which envelope kinds get the error, and greetings are the exception)

kitaebot (approved):
Verdict: **correct** — the drain path does what the commit messages claim, end to end.

Verified against the diff and the surrounding source: the shutdown ordering is exactly drop loops → `begin_drain()` → drop handle → await task, and no loop clones the handle, so the daemon's handle is genuinely the last sender and the actor exits via `recv() → None` after finishing the envelope in hand. The drain check happens at dequeue time, so an envelope queued after the signal is refused rather than executed; the residual window (a send in flight when the socket loop is dropped) resolves to the designed `Err("Agent draining")`, which the socket client actually receives. Greetings are answered state-only with no provider call, so answering them while draining costs nothing. The refusal logic writes the error to the reply channel before dropping the envelope, and `CancellationToken` is the right choice given both sides already depend on the crate.

The systemd reasoning checks out: `KillMode=mixed` SIGTERMs the main process alone and SIGKILLs the cgroup remainder after it exits, so the exec tool's children and MCP servers survive the drained turn, and `TimeoutStopSec=30min` bounds a genuinely hung turn.

The `MockProvider::gated` test pins the behavior deterministically: the yield loop confirms the first turn is in flight before queueing the second, so the observed refusal is the drain logic and not a race, and `call_count() == 1` also proves between-turns compaction was deferred by the refused envelope rather than silently consuming the second response.

One nit, for a follow-up, not blocking:

- `specs/01-agent-loop.md:287` — the spec says the actor "answers every envelope already queued with `Err("Agent draining")`", but greeting envelopes are answered with a greeting while draining (the commit message says so explicitly). Specs are contracts; scope the sentence to queued *input* envelopes.

kitaebot (commented):
Verdict: **incorrect** — the fix replaced the wrong clause and the spec sentence is now garbled.

Commit 7594490 was meant to scope the queued-envelope claim to input envelopes, but it swapped out the *in-flight turn* clause instead. The head state at specs/01-agent-loop.md:287 now reads, in effect: "The actor finishes the queued input envelopes with `Err("Agent draining")` answers every envelope already queued with `Err("Agent draining")` instead of running it..." — two clauses spliced with nothing between them. Two consequences:

1. The spec no longer states the load-bearing behavior this PR exists for: that the in-flight turn finishes (sub-agents awaited inside it). That sentence was the only place the spec said it.
2. The original nit is still unfixed — the surviving clause still says "answers every envelope already queued," which remains false for greeting envelopes.

The intended edit keeps the first clause and touches only the second: e.g. "...finishes the turn it is on (sub-agents are awaited inside the turn, so they finish with it), answers every **input** envelope already queued with `Err("Agent draining")` instead of running it, and exits when the last sender is gone."

Everything else from the first round stands (APPROVE, 7d6c060117): shutdown ordering, refusal logic, systemd semantics, test coverage all verified correct. This is a one-line spec fix away from merge.

specs/01-agent-loop.md
```diff
@@ -281,6 +281,18 @@ hint for display formatting.
 
 If the actor has shut down, `send_message` returns a synthetic error string.
 
+**Shutdown drains.** `AgentHandle::spawn` also returns the actor task's
+`JoinHandle`. On SIGINT/SIGTERM the daemon stops the channel loops, calls
+`begin_drain()`, drops its handle, and awaits the task. The actor finishes the
+queued input envelopes with `Err("Agent draining")`
```
kitaebot, line 287: This edit replaced the wrong clause. It was meant to scope "every envelope already queued" to *input* envelopes; instead it spliced that phrase into the in-flight-turn clause, producing: "The actor finishes the queued input envelopes with `Err("Agent draining")` answers every envelope already queued with `Err("Agent draining")` instead of running it..." — garbled, and it deletes the spec's only statement that the in-flight turn (and its sub-agents) completes.

The intended change keeps the first clause and adds one word to the second:

```suggestion
The actor finishes the turn it is on (sub-agents are awaited inside the turn, so they finish with it), answers every input envelope already queued with `Err("Agent draining")` instead of running it, and exits when the last sender is gone.
```

PR #144, opened 2026-09-02
Re-dispatch reviewer turns a crash interrupted

### LINKED ISSUE #134: Killed tracked-PR reviews never re-dispatch
The PR channel records the reviewed head SHA at dispatch time, deliberately, so a crash mid-review cannot double-review on restart. The cost surfaced 2026-08-28: a review turn killed by a daemon stop left the SHA marked reviewed with no review posted, and the only recovery was amending the commit to move the head (PR #115). The drain work removes the common cause (operator stops), but genuine crashes keep the gap.

Proposal: boot-time reconciliation. On channel start, for each entry in the reviewed map, check whether the bot actually posted a review for that SHA (one API call per tracked PR, and only entries newer than the last completed poll need checking); an entry with no posted review is removed, so the normal pass re-dispatches. Keeps the at-dispatch recording (the double-review fence) while closing the crash loss.

amackillop:
Closes #134.

Both reviewer passes record the head SHA in `reviewed` before the turn runs, so a turn that finishes without a formal review does not re-trigger every tick. The same record swallows a turn the daemon never finished: kill mid-review, SHA stays marked, nothing posted, and the only recovery was amending the head (PR #115).

This does not do the boot-time API reconciliation the issue proposed. GitHub cannot tell a killed turn from the quiet path (publishes nothing on purpose) or from a model that replied without a formal review; an API check would re-dispatch both at every start and defeat the fence. Whether `send()` returned is local state, so that is what gets recorded.

Two commits:

1. `PollState.in_flight` holds the key and the SHA the entry had before dispatch. Written in the same save as the insert, cleared by a second save when the turn returns (Ok or Err). Old state docs deserialize unchanged.
2. At channel start a surviving marker rolls the entry back: key removed for a first review, previous SHA restored for a re-review. Pass 2 or pass 3 then re-dispatches with no new logic. Spec 20 updated (state doc, rollback rule, failure table).

Accepted residue: a kill after the review was submitted but before the second save yields one duplicate review, same class as the corrupt-state case spec 20 already accepts.

Manual check: request a review on a test PR, `kill -9` the daemon while the reviewer turn runs, restart. Expect `Rolling back reviewer turn interrupted by shutdown` in the log and a fresh review dispatch on the next tick.

### DIFF
--- a/specs/20-github.md
+++ b/specs/20-github.md
@@ -544,7 +544,8 @@ replying to stale code.
 The `github_poll` document in the state database (spec 05):
 
 ```json
-{"last_poll": "2026-07-05T12:00:00Z", "reviewed": {"owner/repo#42": "<head-sha>"}}
+{"last_poll": "2026-07-05T12:00:00Z", "reviewed": {"owner/repo#42": "<head-sha>"},
+ "in_flight": {"key": "owner/repo#42", "prev_sha": "<sha-before-dispatch>"}}
 ```
 
 - Missing or corrupt state defaults to `last_poll = now` (avoids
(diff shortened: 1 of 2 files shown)


kitaebot (approved):
Verdict: **correct** — the in-flight marker and boot-time rollback do exactly what the commit messages claim, and the crash-recovery reasoning holds.

Checked against the code, not just the diff: the marker is written in the same single JSON save as the `reviewed` insert in both passes, and cleared by a second save after `send()` returns, Ok or Err (`send` swallows turn errors internally, so there is no miss path). Serde backward compat is right — `#[serde(default)]` plus `skip_serializing_if` means old state docs load unchanged and clean saves don't grow the new field. `roll_back_in_flight` handles the three turn kinds correctly: first review drops the key so the pending request re-dispatches via pass 2, re-review restores the prior SHA so the head-SHA comparison fires pass 3, and a comment-only turn restores the same SHA — matching the pre-existing accepted behavior for comment-only turns. The save-failure corner self-heals through the end-of-tick save, and the accepted residue (kill between review submission and the clear save → one duplicate) is honestly stated and consistent with the corrupt-state case spec 20 already accepts.

The argument against boot-time API reconciliation is sound: the quiet path and a no-formal-review reply both publish nothing, so remote state cannot distinguish a killed turn — `send()` returning is the only real signal, and it is local.

Manual verification of the kill -9 path is the author's claim; the log line and rollback logic it exercises are consistent with the code.

PR #150, opened 2026-09-02
Clamp file_read to an inline window instead of externalizing

### LINKED ISSUE #148: file_read should clamp to an inline window instead of externalizing
## Problem

A `file_read` whose output crosses the engine's inline threshold comes back as an externalized `<file>` reference — a pointer, when the model asked for text it intended to read now. Turn 19 on #136 (2026-09-02) shows the result: whole-file reads of exec.rs (98,935 bytes, ~25k tokens) externalized, while 3-7KB windowed reads came back usable. A model that hits the reference a few times learns "big read = useless" and regresses to `sed -n 'N,Mp'` screenfuls through exec — the prompt's promise that an oversized result "is safe, never lost" is true and beside the point.

Externalizing a file_read is also pure waste: the source file is already on disk at a path the model knows; storing a payload copy adds a second copy of a file, not recoverability.

## Proposal

When the requested range would exceed the inline threshold, return the largest inline-safe window from the requested offset with an explicit continuation line:

    (showing lines 1-400 of 2011; continue with offset=401)

The model gets usable text on every call and the exact incantation for the next slice. No file_read output is ever externalized.

Threshold should come from the same knob the engine uses, not a second constant that drifts.

amackillop:
Closes #148.

A `file_read` above `context.tool_output_tokens` came back as an externalized `<file>` reference — a pointer, where the model asked for text. Turn 19 on #136: 98KB whole-file reads round-tripped into references while 3-7KB windows worked, teaching the model that big reads are useless and pushing it back to `sed -n` screenfuls through exec. Externalizing a file_read is also pure waste; the file is already on disk at a path the model knows.

The tool now takes the engine's inline threshold at construction and returns the largest window that fits, with the trailer extended in place: `(400 lines shown, 2011 total, 98935 bytes; continue with offset=401)`. A 64-token reserve keeps the whole message inline after wrapper framing. One deliberate fall-through: a single line too large to window (minified blob) returns whole and externalizes as before — line windowing can't help it, and the payload store's pretty-printing (#119) is the recovery path.

The continuation rides inside the trailer parentheses because `FILE_READ_TRAILER_RE` requires the trailer as the last line; the regex accepts the optional suffix and the round-trip test covers it. The developer-workflow prompt sentence promising "oversized results externalize safely" now describes the clamp — prompt change, so it takes effect on the next VM rebuild.

Manual verification: `file_read` any 2000-line file with `[context] tool_output_tokens = 500` in config; the result arrives inline with a continuation offset that resumes exactly where it stopped.


src/tools/mod.rs
```diff
@@ -269,7 +269,10 @@ impl Tools {
                 direnv,
                 config.git.trusted_repos(),
             )),
-            Arc::new(FileRead::new(guard.clone())),
+            Arc::new(FileRead::new(
```
kitaebot, line 277: Nit: sub-agent registries (`task.rs` `filtered()`) clone this shared `Arc<FileRead>`, so explore/worker/reviewer agents now clamp at the root's `tool_output_tokens` even though their engine threshold is deliberately 20,000 (`SUB_AGENT_TOOL_OUTPUT_TOKENS`). Recoverable via the continuation offset, but the commit message's "takes the engine's inline threshold" only holds for the root agent. Consider acknowledging it, or wiring the sub-agent threshold through if the 20k absorb-verbose-output design is meant to hold for reads.

kitaebot (approved):
Verdict: **correct** — every mechanism the commit message claims is present and verified in the diff, and the windowing arithmetic is exact.

What I checked against the code: the threshold is genuinely injected from `config.context.tool_output_tokens` (the same field the engine reads for its externalization cutoff), not duplicated. The `take_while` accounting pre-charges each line's real byte cost before the budget check, so the shown lines always fit, and `next = offset + shown` resumes exactly at the first unshown line — pinned by a regression test that round-trips the advertised offset. The 64-token reserve covers the worst-case wrapper + trailer overhead (~15 estimated tokens) with room for the engine's trailing newline. The trailer regex change is backward compatible (suffix optional) and safe for both framing consumers, including the #119 payload normalizer. The minified-single-line fall-through, empty/short files, and a degenerate `< 64` token threshold all behave sanely, and the spec and prompt changes match the implemented behavior.

One nit worth a thought (not blocking): sub-agent tool sets clone this shared `FileRead`, so explore/worker/reviewer agents now clamp at the root's threshold even though their engine threshold is deliberately 20,000 tokens. The continuation offset keeps that functional, but it's an unstated behavior change for sub-agent reads — worth either a sentence in the commit/PR or wiring the sub-agent threshold through if the 20k design is meant to hold.

src/tools/mod.rs
amackillop: Fixed in 68b4f71 rather than acknowledged: the 20k absorb-whole-files design is the point of explore delegation, so clamping sub-agents at 4096 would have degraded the exact path we want the model taking. ToolCtx now carries the per-turn engine threshold (sub-agent child ctx sets SUB_AGENT_TOOL_OUTPUT_TOKENS) and file_read prefers it over the constructed root default. Ctx rather than per-instance state because sub-agent sets are filtered Arc clones. Covered by ctx_threshold_override_lifts_the_clamp.

kitaebot: Confirmed fixed in 68b4f71 — re-reviewed the delta: the ctx override reaches `file_read` on the sub-agent path, the root path is unchanged, and the preference order (ctx wins over the constructed default) is right since the ctx value reflects the engine that will judge the output. Test genuinely varies the knob.

One residual your spec sentence should scope: the distiller (`memory/distill.rs`) also runs a 20k-token `EphemeralSession` but passes `ToolCtx::default()`, so its `file_read` still clamps at the root threshold — the "Sub-agent turns carry their own inline cap" sentence in `specs/03-tools.md` is only true for task-spawned agents. Wiring the distiller's ctx or scoping the sentence would close it; not blocking.

amackillop: Wired rather than scoped: the distiller's ToolCtx now carries DISTILL_TOOL_OUTPUT_TOKENS, same principle as the task path, and the spec sentence names both ephemeral-turn carriers. Pushed.

kitaebot: Confirmed — re-reviewed the delta: distiller ctx now carries `DISTILL_TOOL_OUTPUT_TOKENS` at the same site as the 20k session (shared constant, no drift), every production `EphemeralSession::new` pairing now matches its ctx cap, and the spec sentence names both carriers. Clean.

PR #151, opened 2026-09-02
Inject corrective hints when a failure shape repeats

### LINKED ISSUE #149: Event-triggered corrective hints at the strike and repeat sites
## Problem

Standing prompt guidance does not reach a model mid-habit: developer-workflow.md contains a sentence describing the exact grep/sed pathology it failed to prevent in turn 19 on #136 (2026-09-02), read at iteration 0 and ignored by iteration 90. The midpoint checkpoint (landed 2026-09-02) and the deny-list guidance (#136) both work on the opposite principle — the instruction arrives at the moment its trigger fires. The loop already has the trigger infrastructure: `RepeatDetector`, `ToolStrikeTracker`, per-rule policy strikes.

## Proposal

Append a one-line corrective hint to the tool result when a recognized failure shape fires:

- output contains `assertion left == right failed` — "Derive the expected value by hand from the inputs before changing code; the expectation itself may be wrong."
- identical command re-run with no intervening file edit — "The result will not change until something changes; edit first or ask a different question."
- file_edit no-match for the second time on the same file — "Your model of the file is stale; re-read the region before retrying."

Each hint fires at most once per turn per trigger. Strong-model cost is one line, and only on turns already exhibiting the pathology.

## Design forks

1. Trigger taxonomy and where each lives (exec output scan vs RepeatDetector vs file_edit's error path).
2. Result trailer vs boundary notice (checkpoint precedent says append-only either way, for the cache prefix).
3. Overlap with #145: if test output gets a structured trailer, the assertion hint may belong there instead.
4. Whether hints count toward any strike escalation or stay purely advisory.

amackillop:
Closes #149.

Standing prompt guidance does not reach a model mid-habit — developer-workflow.md describes the grep/sed pathology verbatim and failed to prevent it in both #136 turns. The checkpoint (cb1fea0) established the pattern that works: the instruction arrives when its trigger fires. This applies it per failure shape.

`HintTracker` (new `src/agent/hints.rs`, pure, unit-tested) observes each completed tool call and yields one line per shape, once per turn:

- failed assertion in output → derive the expected value by hand; the expectation may be the bug
- identical test command with no `file_edit`/`file_write` since → the result will not change
- second `file_edit` no-match on one file → your view is stale; re-read the region

Test commands recognized by substring across cargo/just/pytest/pnpm/npm/go/vitest — non-Rust repos are in scope. Hints ride a separate system message after the round's results, not the result text: results are `<tool_output>`-framed and the LCM framing parsers require the closing tag last. Advisory only, no strike coupling; `RepeatDetector` keeps owning literal-identical call sets, while the rerun hint fires earlier on a looser match where the detector correctly stays silent.

Spec 01 gains a "Corrective Hints" section. Loop-level test proves the assert hint lands exactly once as a system notice; six unit tests cover the tracker.

Manual verification: in a session, run the same `just test-one` twice with no edit between — the second result is followed by the rerun hint system message, and a third run stays silent.

### DIFF
--- a/specs/01-agent-loop.md
+++ b/specs/01-agent-loop.md
@@ -214,7 +214,8 @@ tool results are `<tool_output>`-framed and the LCM framing parsers
 require the closing tag last, so appending to the result body would
 break payload normalization. Hints are advisory only; they feed no
 strike counter. The `RepeatDetector` still owns literal-identical
-call sets (three identical rounds skip execution); the rerun hint
+call sets (the third consecutive identical round onward is skipped,
+`REPEAT_LIMIT`); the rerun hint
 fires earlier and on a looser match — same test command, other calls
 allowed in between — where the detector correctly stays silent.
 
--- a/specs/01-agent-loop.md
+++ b/specs/01-agent-loop.md
@@ -192,6 +192,32 @@ directive explicitly says to continue with tool calls: under
 turn summary logs `checkpointed` so effectiveness is measurable
 against `max_iterations` outcomes.
 
+### Corrective Hints
+
+The midpoint checkpoint's principle at finer grain: standing prompt
+guidance does not reach a model mid-habit (developer-workflow.md
+described the grep/sed pathology verbatim and failed to prevent it in
+the #136 turns), so the instruction arrives when its trigger fires.
+Three recognized failure shapes each yield one system-notice line per
+turn, appended after the round's tool results:
+
+| Trigger | Hint |
+|---------|------|
+| Failed assertion in tool output | Derive the expected value by hand; the expectation may be the bug. |
+| Identical test command with no `file_edit`/`file_write` since | The result will not change; edit first. |
+| Second `file_edit` no-match on one file | Your view of the file is stale; re-read the region. |
+
+Test commands are recognized by substring across ecosystems (cargo,
+just, pytest, pnpm/npm, go, vitest): the bot works non-Rust repos.
+The hints ride a separate system message rather than the result text:
+tool results are `<tool_output>`-framed and the LCM framing parsers
+require the closing tag last, so appending to the result body would
+break payload normalization. Hints are advisory only; they feed no
+strike counter. The `RepeatDetector` still owns literal-identical
+call sets (three identical rounds skip execution); the rerun hint
+fires earlier and on a looser match — same test command, other calls
+allowed in between — where the detector correctly stays silent.
+
 ### Budget Exhaustion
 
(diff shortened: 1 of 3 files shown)


specs/01-agent-loop.md
```diff
@@ -192,6 +192,32 @@ directive explicitly says to continue with tool calls: under
...
+just, pytest, pnpm/npm, go, vitest): the bot works non-Rust repos.
+The hints ride a separate system message rather than the result text:
+tool results are `<tool_output>`-framed and the LCM framing parsers
+require the closing tag last, so appending to the result body would
+break payload normalization. Hints are advisory only; they feed no
+strike counter. The `RepeatDetector` still owns literal-identical
+call sets (three identical rounds skip execution); the rerun hint
+fires earlier and on a looser match — same test command, other calls
+allowed in between — where the detector correctly stays silent.
+
```
kitaebot, line 221: Nit: the parenthetical about "three identical rounds" skipping execution is imprecise. `RepeatDetector` (`REPEAT_LIMIT = 3`) executes the first two identical rounds and skips only the third and subsequent consecutive identical round-sets. Given this repo's spec-precision convention, "the third consecutive identical round onward is skipped" would read exactly. Non-blocking.

kitaebot (approved):
Verdict: **correct** — the implementation matches its stated intent on every point the commit message makes, and the injection path is genuinely advisory.

Verified against the code, not the message: the `file_edit` no-match substring matches the actual `ToolError::Precondition` text; failing test output arrives on the exec `Ok` path with the `Exit code: N` trailer, so the assertion scan sees it; the tracker is constructed per turn inside `turn_loop`, so once-per-turn latching is real. Hints ride a bare system message after results — they cannot touch tool results, strike counting, or the `RepeatDetector`, and the separate-message choice is consistent with the `<tool_output>` framing the LCM payload normalization relies on. Six unit tests vary the relevant knobs and the loop test pins exactly-once injection. The tracker's looseness (no path scoping on the edit check, single most-recent-command comparison) is the deliberate design the PR states, with `RepeatDetector` still owning the strict case — accepted.

One nit on spec precision, below; nothing blocking.

specs/01-agent-loop.md
amackillop: Fixed in the follow-up commit: the spec line now reads "the third consecutive identical round onward is skipped (REPEAT_LIMIT)".

kitaebot: Verified against `RepeatDetector::record` (count >= REPEAT_LIMIT, so rounds 1–2 execute, 3+ skip) — the new wording is accurate. Nit closed.

Issue #145, opened 2026-09-02
Structured test evidence: parse libtest output instead of trusting the model's greps


amackillop:
## Problem

The model reads test results through self-authored shell filters and blinds itself. Turn 19 on #136 (2026-09-02) ran `just test-one tools::exec` seven times, each piped through variants of `grep -A2 "panicked" | head -5` — discarding the assertion's left/right values every time — then spent 100+ iterations debugging a test whose expectation one intact panic message would have refuted (300 < 570; the expectation was wrong, not the code). Turn 23 repeated the pattern the same night.

The harness cannot recover what a pipeline inside `bash -c` already discarded, so post-processing the exec result alone is not sufficient.

## Proposal

A structured path for test evidence, in two composable parts:

1. **Recognize unpiped test runs** in exec output (libtest `test result:` lines, `panicked at`, `assertion left == right failed`) and append a mechanical trailer to the tool result: suite counts, failed test names, first assertion's left/right/location. Zero scheduling; the model already runs tests constantly.
2. **A structural exec rule** (same mechanism as the oversized-sleep rule from #136): a command that pipes `cargo test` / `just test*` through a filter gets guidance to run it unpiped — the output is parsed and the full text stays retrievable as an externalized payload.

## Design forks

1. Trailer in the tool result vs boundary notice.
2. Dedicated `test` tool wrapping the recipes with `--format json` instead of pattern-recognizing exec output — cleaner parse, but adds a tool and the model must adopt it.
3. Interaction with externalization: a full test log exceeds the inline threshold; the trailer must survive on the `<file>` stub (see the stub-guidance issue).
4. Recognition scope: libtest only, or nextest/pytest formats for other worked repos.

amackillop:
Scope correction, same as on #103: non-Rust repos are in scope, so fork 4 (recognition formats) is not optional polish — the recognizer should be a small registry from the start. libtest first (all tonight's evidence is Rust), then vitest/jest and pytest, which cover the pnpm monorepos. The structural "don't pipe test commands through filters" rule is already language-agnostic; only the trailer parser is per-format, and an unrecognized format degrades to no trailer, never to a blocked command.

amackillop:
Completed by [PR #153](https://github.com/amackillop/kitaebot/pull/153) (parsed test-evidence trailer) and [PR #167](https://github.com/amackillop/kitaebot/pull/167) (piped-test-command guard).

PR #152, opened 2026-09-02
Name a next step in every externalized file stub

### LINKED ISSUE #147: Externalization stubs should carry a content-aware next step
## Problem

The model never dereferences externalized payloads through the sanctioned paths. Turn 19 on #136 (2026-09-02), 200 iterations: `lcm_grep` 0 calls, `grep` tool 0, `task` (explore) 0. Its substitute for searching the stored copy of a test run is re-running the command — seven identical `just test-one` invocations in one turn. The guidance exists in AGENTS.md ("Large Tool Output") and developer-workflow.md even describes this exact pathology verbatim; static prompt text demonstrably does not transfer mid-habit. The deny-list guidance work (#136) established the pattern that does: instruction at the moment of contact.

## Proposal

`format_file_reference` appends one content-aware line to the `<file>` stub telling the model the next move for that payload type:

- test run: `search it: lcm_grep 'panicked|FAILED' file_xxx`
- source file: `read a window: file_read <path> offset=N limit=M`
- JSON payload: `it is pretty-printed; lcm_grep a key or file_read a range`

Additionally, for recognized formats the summary should lead with a mechanical extract rather than prose: for a test log, the first panic (with assertion left/right) and the counts. Whether the assertion values survive the exploration summary currently depends on the summarizer model's mood; that is the difference between a self-sufficient reference and a forced re-run.

## Design forks

1. Content-type detection: by originating tool + command shape, or by sniffing the payload.
2. Overlap with the structured-test-evidence issue (#145): if exec results get a mechanical trailer at capture time, the stub inherits it for free — decide which layer owns the extract.
3. Stub size budget: the stub replaces an oversized payload; the next-step line and extract must stay well under the inline threshold.

amackillop:
Closes #147.

Models do not dereference `<file>` references through the sanctioned tools unprompted: turn 19 on #136 ran 200 iterations with zero `lcm_grep`/`grep`/`task` calls and re-ran the same test command seven times as its substitute for searching the stored copy. The prose guidance exists in AGENTS.md and demonstrably doesn't transfer mid-habit; the #136 deny-guidance pattern (instruction at the moment of contact) does.

Every stub now ends with a content-aware `Next:` line, chosen by what the payload is:

- test-runner output (cargo/pytest/go shapes) → `lcm_grep 'panicked|FAILED|error' file_xxx`, with an explicit "do not re-run the command"
- payload with an original source path → windowed `file_read <path>`
- structured payloads (JSON/XML/YAML) → the line-oriented stored copy (#119's pretty-printing is why that works)
- everything else → `lcm_grep` the stored text or `file_read` a window

Mechanics: `next_step()` and `looks_like_test_run()` are pure and unit-tested; the engine threads `FileKind`, the source-path flag, and the test-run shape into `format_file_reference`, which appends the line inside the tag body. The summarizer's preserve-`<file>`-tags rule carries it through compaction. An empty next step renders the bare reference, so the compaction/report paths are shape-unchanged. Spec 14's reference example now shows the line.

The mechanical first-panic extract from the issue's fork 2 is deliberately left to #145, which owns capture-time parsing; if that lands, the stub inherits the trailer and this line stays the pointer.

Manual verification: force a small `tool_output_tokens`, exec something producing `test result: FAILED`, and check the stored message's `<file>` stub ends with the lcm_grep instruction.

### DIFF
--- a/src/context/lcm/explore.rs
+++ b/src/context/lcm/explore.rs
@@ -87,17 +87,63 @@ pub fn extract_file_ids(content: &str) -> Vec<String> {
 }
 
 /// Render the `<file>` reference stored in `messages.content` in
-/// place of the raw payload.
+/// place of the raw payload. `next_step` is the content-aware
+/// dereference instruction; empty renders the bare reference.
 pub fn format_file_reference(
     file_id: &str,
     path: &str,
     token_count: usize,
     summary: &str,
+    next_step: &str,
 ) -> String {
-    format!(
-        "<file id=\"{file_id}\" path=\"{path}\" tokens=\"{token_count}\">\n{}\n</file>",
-        summary.trim()
-    )
+    let body = if next_step.is_empty() {
+        summary.trim().to_string()
+    } else {
+        format!("{}\n{next_step}", summary.trim())
+    };
+    format!("<file id=\"{file_id}\" path=\"{path}\" tokens=\"{token_count}\">\n{body}\n</file>")
+}
+
+/// Test-runner output shape, any ecosystem: the payloads a model most
+/// often answers by re-running the command instead of searching.
+pub fn looks_like_test_run(content: &str) -> bool {
+    content.contains("test result:")
+        || content.contains("panicked at")
+        || content.contains("FAILED")
+        || content.contains("=== RUN")
+        || content.contains("passed") && content.contains("failed")
+}
+
+/// The dereference instruction for a `<file>` stub. Models do not
+/// follow references through the sanctioned tools unprompted (a 200
+/// iteration turn issued zero `lcm_grep` calls and re-ran commands
+/// instead), so each stub names its own next step.
+pub fn next_step(
+    kind: FileKind,
+    test_run: bool,
+    has_source_path: bool,
+    file_id: &str,
+    path: &str,
+) -> String {
+    if test_run {
+        return format!(
+            "Next: lcm_grep 'panicked|FAILED|error' {file_id} pulls the \
+             failures from the stored copy; do not re-run the command."
+        );
+    }
+    if has_source_path {
+        return format!("Next: file_read {path} with offset/limit windows the original file.");
+    }
+    match kind {
+        FileKind::Json | FileKind::Xml | FileKind::Yaml => format!(
+            "Next: the stored copy at {path} is line-oriented; lcm_grep a \
+             key ({file_id}) or file_read a range of it."
+        ),
+        _ => format!(
+            "Next: lcm_grep <pattern> {file_id} searches the stored text; \
+             file_read {path} windows it."
+        ),
+    }
 }
 
 /// Reformat a payload for on-disk storage so line-based tools work:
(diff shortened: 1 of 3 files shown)


src/context/lcm/explore.rs
```diff
@@ -87,17 +87,63 @@ pub fn extract_file_ids(content: &str) -> Vec<String> {
...
+pub fn next_step(
+    kind: FileKind,
+    test_run: bool,
+    has_source_path: bool,
+    file_id: &str,
+    path: &str,
+) -> String {
+    if test_run {
+        return format!(
+            "Next: lcm_grep 'panicked|FAILED|error' {file_id} pulls the \
```
kitaebot: This `Next:` line directs the model at `lcm_grep '<pattern>' file_xxx`, but `lcm_grep` has no file-id argument (`GrepArgs` is `pattern`/`mode`/`scope`/`limit`, tools.rs:95) — the trailing id just becomes part of the FTS pattern. Worse, it searches `messages_fts`/`summaries_fts`, and `push_message` persists the message *after* `intercept_large` has already replaced the payload with the stub: the raw bytes exist only on disk under `context/lcm/payloads/`. So the search structurally cannot return the failures it was told to find. The payload is reachable via `lcm_describe file_xxx` / `lcm_expand`, or `file_read` on the already-printed path. The whole point of #147 is guidance that works at the moment of contact; this names the one sanctioned tool that cannot dereference the payload.

kitaebot: Even ignoring the target problem, `panicked|FAILED|error` under the default `fts` mode is an FTS5 syntax error (`no such column: panicked`), and the phrase fallback refuses operator-bearing patterns — the model gets a structured error, not results. The stub needs `mode="regex"` (or a plain-phrase pattern) to have any chance of working.

src/context/lcm/explore.rs
```diff
@@ -87,17 +87,63 @@ pub fn extract_file_ids(content: &str) -> Vec<String> {
...
+    format!("<file id=\"{file_id}\" path=\"{path}\" tokens=\"{token_count}\">\n{body}\n</file>")
+}
+
+/// Test-runner output shape, any ecosystem: the payloads a model most
+/// often answers by re-running the command instead of searching.
+pub fn looks_like_test_run(content: &str) -> bool {
+    content.contains("test result:")
+        || content.contains("panicked at")
+        || content.contains("FAILED")
+        || content.contains("=== RUN")
```
kitaebot, line 112: `content.contains("FAILED")` matches any payload that merely mentions the word — a source file, log excerpt, or document — which then suppresses the correct `file_read`/stored-copy advice in favor of the test-run advice. The cargo (`test result:`/`panicked at`) and pytest (`passed`+`failed` pair) arms already cover the real shapes; this unanchored arm adds false positives for little gain.

src/context/lcm/explore.rs
```diff
@@ -87,17 +87,63 @@ pub fn extract_file_ids(content: &str) -> Vec<String> {
...
+        || content.contains("passed") && content.contains("failed")
+}
+
+/// The dereference instruction for a `<file>` stub. Models do not
+/// follow references through the sanctioned tools unprompted (a 200
```
kitaebot: Nit: this doc comment retells the 200-iteration incident the commit message already carries. Repo style keeps backstory in the commit message; the first sentence (why each stub names its next step) is the part worth keeping.

kitaebot (commented):
Verdict: **incorrect** — the core deliverable, the `Next:` line itself, points at an `lcm_grep` invocation that cannot reach the payload it names.

The wiring is clean otherwise: `next_step`/`looks_like_test_run` are pure and tested, the empty-next-step branch renders the pre-change bytes exactly, and compaction/report shapes are untouched. But the instruction at the moment of contact is the load-bearing part of #147, and it doesn't work as written — details inline.

amackillop:
The review was right and the finding was embarrassing in exactly the way that matters: I recommended a tool invocation without reading the tool's schema, which is the failure class this whole PR series exists to fix. All four findings addressed in the follow-up commit: every Next: line now targets the on-disk copy at the stub's own path via native grep (regex, single-file root works through WalkBuilder) or windowed file_read; the bare FAILED detection arm is dropped with a false-positive regression test; the doc-comment backstory moved to the commit message.

One follow-up outside this PR's scope: AGENTS.md's 'Large Tool Output' section gives the same broken advice ('lcm_grep to search the stored text') — the payload bytes never reach the FTS tables, so that standing prompt line has been wrong since it was written. Worth a small prompt fix on its own.

kitaebot (approved):
Verdict: **correct** — the delta points every `Next:` line at the on-disk copy the stub already names, and the claimed mechanics hold (native `grep` over a single-file root does work through WalkBuilder's file-root yield; the regex alternation is valid under `RegexMatcher`).

All four findings addressed; regression test pins the bare-`FAILED` false positive. One follow-up worth its own issue, as the author flagged: AGENTS.md's 'Large Tool Output' section says "`lcm_grep` to search the stored text" — same broken advice this PR just fixed, wrong since it was written (payload bytes never reach the FTS tables). Out of scope here.

PR #153, opened 2026-09-02
Append parsed test evidence to failing exec runs

amackillop:
Part of #145: the capture-time trailer. The pipe-discouragement half waits on the #136 deny-guidance branch it builds on.

The model reads test results through self-authored filters and discards the decisive lines: turn 19 on #136 piped seven `just test-one` runs through `grep -A2 panicked | head -5` while debugging a wrong expectation whose left/right values any intact panic would have refuted.

Exec output recognized as a failing test run now gets a mechanical trailer at dispatch: summed pass/fail counts across suites, failed test names (first 10), and the first panic block with its assertion values. Recognition is a per-format registry — libtest and pytest today, sized for the non-Rust repos too — degrading to no trailer on unknown output, never an error. Passing runs get nothing: the trailer preserves failure evidence, not green summaries.

The hook sits in `Tools::execute` dispatch, not exec.rs, deliberately: exec.rs has ~200 uncommitted lines in flight on the #136 branch, and derived-from-output content belongs with the output (inside the eventual `<tool_output>` framing, surviving in the mechanical excerpt's tail if the result externalizes).

Review follow-ups (kitaebot, two passes): pytest recognition generalized to any `N category` summary pairs — zero-count omission, error-only, and mixed shapes covered; counts sum across runs; the `Tools::execute` gate has integration tests via a name-parameterized `MockTool`.

Manual verification: exec a failing `just test-one` unpiped; the result ends with `--- parsed test evidence ---` carrying counts, names, and the left/right values.

### DIFF
--- a/src/tools/mod.rs
+++ b/src/tools/mod.rs
@@ -435,14 +435,18 @@ mod tests {
     use super::*;
     use crate::types::ToolFunction;
 
-    fn mock_call(id: &str) -> ToolCall {
+    fn call_named(name: &str, id: &str) -> ToolCall {
         ToolCall::new(
             id.to_string(),
             ToolFunction {
-                name: "mock".parse().unwrap(),
+                name: name.parse().unwrap(),
                 arguments: "{}".to_string(),
             },
         )
+    }
+
+    fn mock_call(id: &str) -> ToolCall {
+        call_named("mock", id)
     }
 
     #[derive(serde::Deserialize)]
@@ -584,22 +588,12 @@ mod tests {
 
     const FAILING_RUN: &str = "test a ... FAILED\n\ntest result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s\n";
 
-    fn named_call(name: &str) -> ToolCall {
-        ToolCall::new(
-            "c1".to_string(),
-            ToolFunction {
-                name: name.parse().unwrap(),
-                arguments: "{}".to_string(),
-            },
-        )
-    }
-
     /// The dispatch hook (issue #145) fires only for exec results.
     #[tokio::test]
     async fn exec_failing_test_output_gets_the_evidence_trailer() {
         let tools = Tools::new(vec![Arc::new(MockTool::named("exec", FAILING_RUN))], &[]).unwrap();
         let out = tools
-            .execute(&named_call("exec"), ToolCtx::default())
+            .execute(&call_named("exec", "c1"), ToolCtx::default())
             .await
             .unwrap();
         assert!(out.contains("--- parsed test evidence ---"), "{out}");
@@ -611,7 +605,7 @@ mod tests {
     async fn non_exec_tool_output_is_untouched() {
         let tools = Tools::new(vec![Arc::new(MockTool::new(FAILING_RUN))], &[]).unwrap();
         let out = tools
-            .execute(&named_call("mock"), ToolCtx::default())
+            .execute(&call_named("mock", "c1"), ToolCtx::default())
             .await
             .unwrap();
         assert_eq!(out, FAILING_RUN);
@@ -624,7 +618,9 @@ mod tests {
             &[],
         )
         .unwrap();
-        let result = tools.execute(&named_call("exec"), ToolCtx::default()).await;
+        let result = tools
+            .execute(&call_named("exec", "c1"), ToolCtx::default())
+            .await;
         assert!(matches!(result, Err(ToolError::Blocked { .. })));
     }
 
--- a/src/tools/mod.rs
+++ b/src/tools/mod.rs
@@ -582,6 +582,52 @@ mod tests {
         ));
     }
 
+    const FAILING_RUN: &str = "test a ... FAILED\n\ntest result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s\n";
+
+    fn named_call(name: &str) -> ToolCall {
+        ToolCall::new(
+            "c1".to_string(),
+            ToolFunction {
+                name: name.parse().unwrap(),
+                arguments: "{}".to_string(),
+            },
+        )
+    }
+
+    /// The dispatch hook (issue #145) fires only for exec results.
+    #[tokio::test]
+    async fn exec_failing_test_output_gets_the_evidence_trailer() {
+        let tools = Tools::new(vec![Arc::new(MockTool::named("exec", FAILING_RUN))], &[]).unwrap();
+        let out = tools
+            .execute(&named_call("exec"), ToolCtx::default())
+            .await
+            .unwrap();
+        assert!(out.contains("--- parsed test evidence ---"), "{out}");
+        assert!(out.contains("0 passed, 1 failed"), "{out}");
+        assert!(out.starts_with(FAILING_RUN), "original output must lead");
+    }
+
+    #[tokio::test]
+    async fn non_exec_tool_output_is_untouched() {
+        let tools = Tools::new(vec![Arc::new(MockTool::new(FAILING_RUN))], &[]).unwrap();
+        let out = tools
+            .execute(&named_call("mock"), ToolCtx::default())
+            .await
+            .unwrap();
+        assert_eq!(out, FAILING_RUN);
+    }
+
+    #[tokio::test]
+    async fn exec_errors_pass_through_the_evidence_hook() {
+        let tools = Tools::new(
+            vec![Arc::new(MockBlockedTool::named("exec", "denied"))],
+            &[],
+        )
+        .unwrap();
+        let result = tools.execute(&named_call("exec"), ToolCtx::default()).await;
+        assert!(matches!(result, Err(ToolError::Blocked { .. })));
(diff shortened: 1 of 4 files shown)


src/tools/test_evidence.rs
```diff
@@ -0,0 +1,187 @@
...
+
+static LIBTEST_RESULT_RE: LazyLock<Regex> = LazyLock::new(|| {
+    crate::text::static_regex(r"(?m)^test result: (ok|FAILED)\. (\d+) passed; (\d+) failed;")
+});
+
+static LIBTEST_FAILED_RE: LazyLock<Regex> =
+    LazyLock::new(|| crate::text::static_regex(r"(?m)^test (\S+) \.\.\. FAILED$"));
+
+static PYTEST_RESULT_RE: LazyLock<Regex> =
+    LazyLock::new(|| crate::text::static_regex(r"(?m)^=+ .*?(\d+) failed, (\d+) passed.*? =+$"));
```
kitaebot: This regex requires the adjacent literal `(\d+) failed, (\d+) passed`, but pytest omits zero-count categories from its summary line: an all-fail run prints `=== 3 failed in 0.12s ===` with no `passed` count at all — no match, no trailer. That is exactly the single-failing-test debugging run this feature exists for (the pytest analogue of the `just test-one` incident in the commit message). Error-only summaries (`1 failed, 1 error in ...`) miss the same way, since `passed` must immediately follow `failed, `. The passing-run test only exercises libtest green output, so nothing catches this.

```suggestion
static PYTEST_RESULT_RE: LazyLock<Regex> =
    LazyLock::new(|| crate::text::static_regex(r"(?m)^=+ .*?(\d+) failed(?:, (\d+) passed)?.*? =+$"));
```

kitaebot (commented):
Verdict: **incorrect** — the pytest recognition regex misses the all-fail summary pytest prints for exactly the single-failing-test debugging run this feature exists for.

The placement in `Tools::execute` is sound: exec returns `Ok` with the rendered output plus `Exit code: N` even on non-zero exits, so the hook does see failing runs, and the `<tool_output>` framing wraps downstream, so the trailer lands inside it and survives the mechanical excerpt's tail. The libtest side — multi-suite summing, name truncation, panic-block extraction — matches the spec paragraph, and I could not construct a realistic false positive on ordinary build or git output.

The must-fix is the one gap: `PYTEST_RESULT_RE` requires the adjacent literal `(\d+) failed, (\d+) passed`, but pytest omits zero-count categories from its summary line. An all-fail run prints `=== 3 failed in 0.12s ===` with no `passed` count at all — no match, no trailer. That is the pytest analogue of the `just test-one` incident in the commit message.

src/tools/test_evidence.rs
```diff
@@ -0,0 +1,187 @@
...
+        failed_names: LIBTEST_FAILED_RE
+            .captures_iter(output)
+            .map(|c| c[1].to_string())
+            .collect(),
+    })
+}
+
+/// Pytest short summary: `=== 2 failed, 5 passed in 0.31s ===`.
+fn pytest(output: &str) -> Option<Evidence> {
+    let cap = PYTEST_RESULT_RE.captures(output)?;
```
kitaebot: libtest() sums counts across every `test result:` line, but this uses `captures()` — first match only — while the FAILED-name list below spans the whole output. `pytest unit/ && pytest integration/` in one exec yields counts from the first run and names from both. Iterate and sum, matching the sibling format.

src/tools/mod.rs
```diff
@@ -299,7 +300,16 @@ impl Tools {
     pub async fn execute(&self, call: &ToolCall, ctx: ToolCtx) -> Result<String, ToolError> {
         tracing::debug!(tool = %call.function.name, call_id = %call.id, "Tool started");
         let started = std::time::Instant::now();
-        let result = self.dispatch(call, ctx).await;
+        let mut result = self.dispatch(call, ctx).await;
+        // Exec output recognized as a failing test run gets the parsed
+        // verdict appended, surviving whatever the command's own
+        // pipeline filtered away (issue #145).
+        if call.function.name.as_str() == "exec"
```
kitaebot, line 307: The `Tools::execute` hook — the only integration claim of this PR — is untested: all six tests call `test_evidence::trailer` directly, so nothing pins the `name == "exec"` gate, the Ok-only gate, or that non-exec tools and `Err` results pass through unmodified. `MockTool` already exists in this module's test suite; a mock named `exec` returning failing-test output would cover it.

amackillop:
All three findings were valid; fixed in the follow-up commit. Pytest passed-group is optional and counts sum across runs via captures_iter (two new tests: all-fail summary, multi-run sum). The dispatch hook has integration coverage through a name-parameterized MockTool: trailer on exec, untouched non-exec, errors pass through. Gate green.

src/tools/test_evidence.rs
```diff
@@ -0,0 +1,217 @@
...
+
+static LIBTEST_RESULT_RE: LazyLock<Regex> = LazyLock::new(|| {
+    crate::text::static_regex(r"(?m)^test result: (ok|FAILED)\. (\d+) passed; (\d+) failed;")
+});
+
+static LIBTEST_FAILED_RE: LazyLock<Regex> =
+    LazyLock::new(|| crate::text::static_regex(r"(?m)^test (\S+) \.\.\. FAILED$"));
+
+static PYTEST_RESULT_RE: LazyLock<Regex> = LazyLock::new(|| {
+    crate::text::static_regex(r"(?m)^=+ .*?(\d+) failed(?:, (\d+) passed)?.*? =+$")
```
kitaebot: The optional group fixes the all-fail case, but the regex still hard-codes the `failed`/`passed` category words: pytest summaries naming other categories — the `1 failed, 1 error in ...` shape from the original finding, or an all-error `1 error in 0.12s` run with zero failed — either silently drop the error count or get no trailer at all. The zero-count-omission problem is only partially closed; consider matching any `N <category>` pairs and summing the non-passed ones.

src/tools/mod.rs
```diff
@@ -572,6 +582,52 @@ mod tests {
         ));
     }
 
+    const FAILING_RUN: &str = "test a ... FAILED\n\ntest result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.01s\n";
+
+    fn named_call(name: &str) -> ToolCall {
```
kitaebot: `named_call(name)` subsumes the pre-existing `mock_call` helper further up this test module, which builds the same `ToolCall` with a hardcoded `"mock"` name. Two helpers to build the same thing in one module; widen the first instead.

kitaebot (commented):
Verdict: **incorrect** — all three prior findings are addressed as claimed, but the pytest fix stops at `failed`/`passed` and the error-only summary shapes from the original finding still parse wrong or not at all.

The optional passed group is the right fix and holds under real pytest ordering; `cap.get(2).map_or(0, ...)` handles the absent group; the three integration tests genuinely pin the exec gate, non-exec pass-through, and error pass-through; `pytest()` now sums like its libtest sibling with a two-run test. The residual is the same class as the must-fix, one level down: a summary naming only other categories (`1 failed, 1 error in ...` drops the error count; `1 error in 0.12s` with zero failed gets no trailer) — sub-blocking, since the all-fail case that motivated the feature now works.

amackillop:
Both residuals fixed. Pytest parsing now scans every 'N category' pair on a summary bar and classifies: passed counts as passed, failed/error(s) count as failures (a collection error is a failure for evidence purposes), everything else ignored — error-only, mixed, and warnings-only shapes covered by tests, and section headers like '=== FAILURES ===' carry no known category so they cannot mark plain output as a pytest run. The two call-building helpers are unified into call_named with mock_call as a thin wrapper. Gate green.

src/tools/mod.rs
```diff
@@ -425,16 +435,20 @@ mod tests {
...
             id.to_string(),
             ToolFunction {
-                name: "mock".parse().unwrap(),
+                name: name.parse().unwrap(),
                 arguments: "{}".to_string(),
             },
         )
     }
 
+    fn mock_call(id: &str) -> ToolCall {
```
kitaebot, line 448: Same duplicate-helper class as the prior nit, one step short: `mock_call` is now a one-line forwarder to `call_named` with exactly one remaining call site (`test_execute`). Inlining `call_named("mock", "test-123")` there and deleting the forwarder finishes the fold. Nit only.

kitaebot (approved):
Verdict: **correct** — both residuals from the prior round are fixed; one nit remains on the helper fold.

The category-scanning parser closes the zero-count hole completely: all-fail, error-only, mixed-category, warnings-only, and section-header shapes all classify correctly and are pinned by tests, and multi-run summation holds. The whitelist guard holds against false positives — generic `==== 5 done ====`-style bars and cargo/gradle summary lines carry no whitelisted category word, so they can't mark plain output as a pytest run; `=== FAILURES ===` headers can't either. The pytest-raises parenthesized variant (`(RAISES mode)`) degrades to no trailer, which is the designed unknown-format behavior. The helper unification serves all new call sites; the only residue is that `mock_call` survives as a forwarder for one caller (inline).

PR #155, opened 2026-09-02
Guide blocked commands and deny certain-timeout sleeps

### LINKED ISSUE #136: Deny-list guidance: 63 of 92 rules say only 'command blocked by policy'

amackillop:
Closes #136.

Three commits, one concern each.

**1. Stop striking redirects to stream devices.** The journal from 2026-08-31 to 2026-09-02 has six "command blocked by policy" strikes on plain `> /dev/null` / `>> /dev/null` redirects. The device-write rule (`>\s*/dev/`) was meant for `> /dev/sda`; commit 30b1d98 only excluded the `2>` form. The "Please advise how to proceed" halt the issue attributes to a `/tmp` reference was this: `cat /tmp/commit-119-nits.diff >> /dev/null` matched on the `>> /dev/null`, not the path. The stream devices (null, zero, full, std*, tty, fd/N, urandom) are now blanked before the regex layer, the same way payload-store reads already are. `> /dev/sda` still blocks.

**2. Guidance on every rule with a remediation.** 39 rules move off the bare message onto shared per-class constants (recursive rm, chmod, ownership, host power, kernel, signals, scheduling, shred/wipe, truncate, mounts, nix mutations), plus reworded nix guidance. Exfiltration, pipe-to-shell, reverse shells, recon, secret/keyring harvesting, LD_PRELOAD, nsenter and firewall stay bare, since any hint there is a how-to. Spec 03 gets a guidance-policy paragraph.

**3. Deny sleeps that cannot fit the budget.** Structural rule: a `sleep` in command position whose literal arguments (GNU suffixes, fractions, `infinity`) sum past `timeout_secs - 30s` is denied with guidance naming the budget, the longest poll that fits, and `github_ci_status`. `sleep $VAR` is left to the shell. The verdict type becomes `Cow<'static, str>` to carry the numbers; the budget comes from the tool's configured timeout.

Not done, deliberately: the issue asks for `/tmp` and raw-`cargo` guidance. Neither has a deny rule, and the log lines behind both were the `/dev/null` false positive in commit 1. Adding a cargo rule would break the daemon on every non-just repo.

Verify beyond the unit tests, after a rebuild:
- `just ask` an exec of `git status > /dev/null && echo ok` (runs, no strike).
- `sleep 600; echo waited` (blocked, message names 600s / 570s / github_ci_status).
- `rm -rf foo` (blocked with the single-file rm hint).

### DIFF
--- a/specs/03-tools.md
+++ b/specs/03-tools.md
@@ -135,7 +135,8 @@ Executes commands via `bash -c` within the workspace.
    model and the sandbox grants; redirection into `state/`),
    disk/filesystem (`mkfs`, `dd if=`, `fdisk`, redirects into `/dev/`
    other than the stream devices `/dev/null`, `/dev/std*`, `/dev/fd/N`,
-   `/dev/tty`, `/dev/zero`, `/dev/full`, `/dev/urandom`), system power (`init 0-6`,
+   `/dev/tty`, `/dev/zero`, `/dev/full`, `/dev/random`, `/dev/urandom`),
+   system power (`init 0-6`,
    `systemctl`), privilege escalation (`sudo`, `chmod`, `chown`),
    network exfiltration (`curl -T`, `nc -l`, `socat`), pipe-to-shell
    (`curl|sh`, `wget|sh`), reverse shells (`/dev/tcp/`, python/ruby/perl
@@ -179,7 +180,8 @@ raw git verbs to the git tools, nix mutations to "the operator deploys",
 signal sweeps to the automatic child cleanup). Only classes where any
 hint would be a how-to keep the bare `command blocked by policy`:
 exfiltration, pipe-to-shell, reverse shells, recon, secret and keyring
-harvesting, library injection, namespace escape, firewall tampering.
+harvesting, library injection, namespace escape, firewall tampering,
+fork bombs.
 
 **Environment scrubbing:**
 
--- a/specs/03-tools.md
+++ b/specs/03-tools.md
@@ -165,6 +165,15 @@ Executes commands via `bash -c` within the workspace.
 These are defense-in-depth heuristics with friendly error messages. The real
 filesystem boundary is the Landlock sandbox (see [spec 15](15-sandbox.md)).
 
+**Guidance policy** (issue #136): the block message is what the model
+reads after a strike, so every rule class with a remediation states it
+(recursive rm to single-file rm, chmod +x to running the interpreter,
+raw git verbs to the git tools, nix mutations to "the operator deploys",
+signal sweeps to the automatic child cleanup). Only classes where any
+hint would be a how-to keep the bare `command blocked by policy`:
+exfiltration, pipe-to-shell, reverse shells, recon, secret and keyring
+harvesting, library injection, namespace escape, firewall tampering.
+
 **Environment scrubbing:**
 
 Child processes run with only allowlisted env vars forwarded:
--- a/specs/03-tools.md
+++ b/specs/03-tools.md
@@ -133,7 +133,9 @@ Executes commands via `bash -c` within the workspace.
    (workspace-root `context/` references except reads of
    `context/lcm/payloads/<file_id>`, which `<file>` references hand to the
    model and the sandbox grants; redirection into `state/`),
-   disk/filesystem (`mkfs`, `dd if=`, `fdisk`), system power (`init 0-6`,
+   disk/filesystem (`mkfs`, `dd if=`, `fdisk`, redirects into `/dev/`
+   other than the stream devices `/dev/null`, `/dev/std*`, `/dev/fd/N`,
+   `/dev/tty`, `/dev/zero`, `/dev/full`, `/dev/urandom`), system power (`init 0-6`,
    `systemctl`), privilege escalation (`sudo`, `chmod`, `chown`),
    network exfiltration (`curl -T`, `nc -l`, `socat`), pipe-to-shell
    (`curl|sh`, `wget|sh`), reverse shells (`/dev/tcp/`, python/ruby/perl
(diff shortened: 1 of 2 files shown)


specs/03-tools.md
```diff
@@ -133,7 +133,9 @@ Executes commands via `bash -c` within the workspace.
    (workspace-root `context/` references except reads of
    `context/lcm/payloads/<file_id>`, which `<file>` references hand to the
    model and the sandbox grants; redirection into `state/`),
-   disk/filesystem (`mkfs`, `dd if=`, `fdisk`), system power (`init 0-6`,
+   disk/filesystem (`mkfs`, `dd if=`, `fdisk`, redirects into `/dev/`
+   other than the stream devices `/dev/null`, `/dev/std*`, `/dev/fd/N`,
```
kitaebot, line 137: Nit (specs-are-contracts): the spared-device list names `/dev/urandom`, but the code also blanks `/dev/random` — the `u?random` alternative in `STREAM_DEVICE_RE` (`src/tools/exec.rs:477`). Either add `/dev/random` to this list or narrow the regex to `urandom` only.

specs/03-tools.md
```diff
@@ -158,11 +160,27 @@ Executes commands via `bash -c` within the workspace.
...
+   `sleep 600; echo waited` timeouts on 2026-09-01). GNU suffixes,
+   fractions, and `infinity` are read; a variable argument is left to
+   the shell. The block names the budget, the longest poll interval
+   that fits, and `github_ci_status` for CI waits.
 
 These are defense-in-depth heuristics with friendly error messages. The real
 filesystem boundary is the Landlock sandbox (see [spec 15](15-sandbox.md)).
 
+**Guidance policy** (issue #136): the block message is what the model
+reads after a strike, so every rule class with a remediation states it
```
kitaebot, line 177: Nit: this paragraph lists the classes keeping the bare `command blocked by policy` message as if exhaustive, but omits the fork-bomb rule (`src/tools/exec.rs:313`), which still uses `BLOCKED`. Add it to the list or give it guidance.

kitaebot (approved):
Verdict: **correct** — the three mechanisms hold up under hand-tracing: stream-device blanking preserves the block-device rule's purpose, the sleep budget rule handles the claimed argument forms with the right boundary, and the guidance constants are attached to exactly the classes that have a remediation.

I hand-traced the diff and the head state of `src/tools/exec.rs`:

- **Stream-device blanking (commit 1):** the regex blanks only the closed stream set, `\b` prevents over-blanking look-alikes (`/dev/tty1`, `/dev/nullify` stay blocked), and `> /dev/sda` / `>/dev/nvme0n1` blocking is asserted. The 30b1d98 `2>` exclusion remains via the blanking path. No previously-blocked harmful command passes — the only rule keyed on `/dev/` text is the device-write one.
- **Sleep rule (commit 3):** `sleep_arg_secs` handles plain numbers, suffixes, fractions, `infinity`; negatives are excluded twice over; bare `sleep` sums to 0 and passes; `sleep $VAR` and arithmetic fail parse and reach the shell, so no false strikes on polling loops. Boundary is `570` allowed / `571` blocked at the 600s default. The `Cow` change is sound end-to-end, and the reworked timeout tests genuinely reach the timeout path.
- **Guidance constants (commit 2):** spot-checked rm/chmod/kill/truncate/shred/nix mappings and confirmed the how-to classes stay bare, matching the new spec paragraph.

Two spec nits, both in text this PR added — under specs-are-contracts they should not drift:

1. `specs/03-tools.md:137` — the spared-device list names `/dev/urandom` but the code also blanks `/dev/random` (`u?random` in `STREAM_DEVICE_RE`). Add `/dev/random` to the list or narrow the regex.
2. `specs/03-tools.md:176` — the paragraph listing classes that keep the bare message presents itself as exhaustive but omits the fork-bomb rule, which still uses it.

Neither affects runtime behavior. The deliberately-not-done /tmp and cargo items are correctly reasoned: no deny rule exists to guide on, and a cargo rule would break on every non-just repo.

[ledger: pending]

specs/03-tools.md
amackillop: Fixed in f49de63.

specs/03-tools.md
amackillop: Fixed in f49de63.

specs/03-tools.md
kitaebot: f49de63 — confirmed fixed. Head-state regex is `u?random` and the spec list now names both `/dev/random` and `/dev/urandom`, so spec and code agree; narrowing the regex would have reintroduced the false-strike class commit 1 removed, so widening the spec was the right side to fix.

specs/03-tools.md
kitaebot: f49de63 — confirmed fixed. Audited the head-state `DENY_RULES` table: fork bombs are the only rule beyond the prior eight still on the bare `BLOCKED` message, so the list is now exhaustive as claimed.

PR #162, opened 2026-09-17
Search externalized payloads through LCM

### LINKED ISSUE #154: AGENTS.md tells the model to lcm_grep payloads the FTS tables never contain
## Problem

The "Large Tool Output" section of AGENTS.md advises: "`lcm_grep` to search the stored text for the part you need." That advice cannot work: `intercept_large` replaces oversized payloads with the `<file>` stub before persistence, so the raw bytes exist only on disk under `context/lcm/payloads/` — `lcm_grep` searches `messages_fts`/`summaries_fts` (scope: messages, summaries, both) and structurally cannot see them. Surfaced by kitaebot's review of PR #152, which caught the same broken advice being baked into the stub next-step line.

A model that follows the standing prompt gets zero results (or an FTS5 syntax error for operator-bearing patterns) and learns that dereferencing payloads is useless — the exact re-run-the-command pathology #147 exists to fix, taught by our own prompt.

## Proposal

Fix the AGENTS.md bullet to name the paths that work: native `grep` against the stub's `path` attribute, windowed `file_read`, `lcm_describe`/`lcm_expand` by file id, or explore delegation. One prompt edit; takes effect on the next VM rebuild.

Alternatively, if payload search via `lcm_grep` is *wanted*, that's a feature: a `files` scope that greps the payload store. But the prompt must not promise it before it exists.

amackillop:
Closes #154.

Adds `scope: "files"` to `lcm_grep`. File searches use regex over the immutable LCM payload copy, so mutable workspace paths cannot affect retrieval. FTS remains limited to messages and summaries.

Validation: `just check`.


src/context/lcm/tools.rs
```diff
@@ -268,6 +298,60 @@ fn run_grep(
...
+    pattern: &str,
+    limit: u32,
+    payloads_dir: &std::path::Path,
+) -> Result<Vec<String>, ToolError> {
+    let matcher = RegexMatcher::new(pattern)
+        .map_err(|e| ToolError::InvalidArguments(format!("invalid regex: {e}")))?;
+    let mut stmt = conn
+        .prepare(
+            "SELECT file_id FROM large_files \
+             WHERE conversation_id = ?1 ORDER BY created_at, file_id",
```
kitaebot, line 313: Must-fix: the `UTF8` sink errors on invalid UTF-8 in any matched line (grep-searcher docs: "All matches must be UTF-8 … an error is returned and searching stops. Use the `Lossy` sink instead"), and that error propagates out of `grep_files` via `?` — so one non-UTF-8 match line in one payload aborts the entire file search. Oversized tool-output payloads are exactly where mixed/binary-ish encodings show up.

```suggestion
                UTF8(|line, text| {
```

should be the `Lossy` sink (same closure shape, lossily substitutes invalid bytes instead of aborting):

```suggestion
                Lossy(|line, text| {
```

with `use grep::searcher::sinks::Lossy;`. Alternatively keep `UTF8` and map per-file errors to a skip, but `Lossy` is the one-line fix and the snippet is for humans anyway.

kitaebot (commented):
Verdict: **incorrect** — one real robustness gap in the new file-search path; everything else checks out.

**Note on process:** the `reviewer` sub-agent could not run — four attempts over ~15 minutes all failed with a provider-side 402 (`in_flight_budget_exhausted`, honored the Retry-After windows). Per protocol the gate review was skipped and I judged the diff myself. Unverified-by-second-pair caveat applies.

What I checked:

- The commit message's central claim holds: `grep_files` enumerates `large_files` rows for the active conversation and searches `payloads_dir.join(file_id)` — the immutable LCM payload copy — never the mutable `path` column. The test `lcm_grep_regex_mode_files_reads_authoritative_payload` pins exactly this (source file contains a match that must not appear).
- Path safety: `extract_file_ids(&file_id).into_iter().next().is_some_and(|id| id == file_id)` rejects malformed IDs from `large_files` before they reach `PathBuf::join`; a bad row is a hard `Precondition` error, not silent path traversal.
- `scope: "files"` + non-regex mode is rejected at argument validation before any work, with a clear message, and is covered by a test.
- Spec (specs/14-context-engine.md) and src/prompts/AGENTS.md updates match the implementation, including the regex-only constraint.
- Tool description updated consistently.

Findings inline below. The must-fix is the `UTF8` sink: per the vendored grep-searcher 0.1.16 docs, "All matches must be UTF-8 … If this sink sees invalid UTF-8, then an error is returned and searching stops." That error propagates out of `grep_files` via `?`, so a single non-UTF-8 match line — entirely plausible in oversized tool-output payloads (binary-ish output, mixed encodings) — aborts the whole file search instead of skipping the line. The `Lossy` sink exists for exactly this and changes nothing else about the result shape.

No ledger ids accompany this review (recording failed on the reviewer side); findings are published in prose only.

src/context/lcm/tools.rs
amackillop: Not applicable. LCM payloads are written only from `&str`: `normalize_payload` returns `Cow<str>`, and `externalize` writes that value. `lcm_expand` also reads the same files with `read_to_string`, so valid payload storage is UTF-8 by contract. A corrupt on-disk payload should surface as a structured I/O error, not be silently rewritten with replacement characters. Keeping `UTF8`.

kitaebot: Conceded — you're right, and the code backs the contract: `externalize` is the only writer to `payloads_dir` and it takes `content: &str`, writing `normalize_payload(content).as_ref()` (engine.rs ~321). Raw tool bytes never reach a payload file, so a non-UTF-8 match line means on-disk corruption, and a structured I/O error is the correct response there — silent replacement characters would mask it. `Lossy` would have been the wrong call; my premise (binary-ish encodings surviving into stored payloads) doesn't hold. Keeping `UTF8` is correct.

PR #165, opened 2026-09-18
Record compaction provider usage

### LINKED ISSUE #139: Summarizer and raw-chat provider calls never reach the usage ledger

amackillop:
Closes #139.

Records successful and failed summarizer/cache-prefix provider calls as standalone background ledger rows. Between-turn compaction cannot be attributed reliably to a later envelope.

Validation: `just test-one context::tests::summarize_fn_records_successful_usage`, `just test-one context::tests::raw_chat_records_usage`, `just test-one summarize_fn_records_failed_attempt_usage`, `just check`.


src/usage.rs
```diff
@@ -218,6 +236,60 @@ impl UsageLedger {
         )?;
         Ok(())
     }
+
+    fn record_provider_call(&self, record: &ProviderCallRecord<'_>) -> rusqlite::Result<()> {
+        let calls = record.calls.iter().cloned().fold(
+            (0u32, 0u64, None, 0u64, None, None),
+            |(count, prompt, cached, completion, cost, provider), call| {
+                (
+                    count + 1,
```
kitaebot: Nit: this fold re-implements `TurnUsage::add_call`'s per-call summing (src/agent/mod.rs:297) as a raw seven-tuple instead of reusing that accumulator. The `Some`-tracking on cached/cost and `or(provider)` newest-endpoint rules now live in two places and can drift when the aggregation rule next changes. `TurnUsage` is `pub(crate)` and constructible element-by-element, so reuse is feasible — or at minimum acknowledge the duplication.

kitaebot (approved):
Verdict: **correct** — the change does exactly what the commit message claims, and the between-turn-unattributable rationale holds up against the code.

Summarizer and cache-prefix calls now write `background:summarizer` / `background:cache-prefix` ledger rows on both the success and failure paths, with failed draws folded first — matching how `TurnUsage::add_failed` settles #128 billing. The unattributability claim is real: `compact_between_turns` and the `compact_if_urgent` paths fire after `record_turn` has already written the turn's row, so charging compaction cost to a later envelope would misstate task spend. The spec-27 edit states the policy in its proper home, the shared `usage_ledger()` preconstruction at startup reaches both engine wiring paths (LCM and flat) identically, and existing call sites are updated to the ledger-absent `None` variant. Tests cover the three claimed validations.

One nit, non-blocking: `UsageLedger::record_provider_call`'s seven-tuple fold re-implements `TurnUsage::add_call`'s per-call summing (src/agent/mod.rs:297) instead of reusing that accumulator — the `Some`-tracking on cached/cost and newest-provider-wins rules now live in two places and can drift when the aggregation rule next changes.

src/context/mod.rs
```diff
@@ -278,13 +284,47 @@ pub fn make_summarize_fn<P: Provider + 'static>(provider: Arc<P>) -> SummarizeFn
...
+                .as_secs();
+            let task = TaskKey::background("summarizer");
+            let outcome = provider.chat("summarizer", &prompt_messages, &[]).await;
+            match &outcome {
+                Ok(outcome) => usage::record_provider_call(
+                    ledger.as_deref(),
+                    &ProviderCallRecord {
+                        calls: outcome
+                            .failed
+                            .iter()
```
kitaebot: Nit: this builder constructs `ProviderCallRecord` as a full Ok/Err literal pair, and `make_raw_chat_fn` (src/context/mod.rs:~364) does the same — the provider-side record shape (all failed calls plus, on success, the final call collected into one `Vec`) is spelled out twice per builder. A small helper taking the outcome-side calls plus duration/model/session/source would collapse each builder to a single assembly. Left as a nit since the aggregation itself is now shared via `add_call`.

src/context/mod.rs
```diff
@@ -312,15 +352,56 @@ pub type RawChatFn = Arc<
...
+                match &outcome {
+                    Ok(outcome) => usage::record_provider_call(
+                        ledger.as_deref(),
+                        &ProviderCallRecord {
+                            calls: outcome
+                                .failed
+                                .iter()
+                                .cloned()
+                                .chain(std::iter::once(outcome.usage.clone()))
+                                .collect(),
```
kitaebot: Nit: same Ok/Err literal duplication as `make_summarize_fn` (see comment there) — a shared record-assembly helper would collapse this to a single construction per builder.

kitaebot (commented):
Verdict: **correct** — the delta addresses the pending duplicate-helper finding cleanly and introduces no regressions.

`record_provider_call` now folds `CallUsage` values through `TurnUsage::add_call` (widened to `pub(crate)`), so the some-tracking and provider-newest-wins aggregation rules live in one place again; the raw seven-tuple became the typed `ProviderCallRecord`. Success-path recording chains failed draws before the final call on the pre-`?` outcome and the Err arm records before returning, matching the turn loop's record-then-`?` pattern. Tests are parameterized away from pre-change literals and cover success and failure on the `SummarizeFn` side.

One leftover nit, non-blocking (new finding): the two builders each hand-assemble a `ProviderCallRecord` twice as full Ok/Err literal pairs — the provider-side shape is spelled out twice per builder where a small helper would collapse each to one assembly. Aggregation itself is shared now, so this is the record-assembly boundary only.

PR #166, opened 2026-09-18
Supervise unexpected actor exits

### LINKED ISSUE #156: A panicking turn kills the agent actor permanently; daemon keeps polling with no agent
## Evidence

On 2026-08-30T06:20:42Z the distill duty applied a `file_edit` to a memory file and the tool panicked:

```
panic: panicked at src/tools/file_edit.rs:406:12:
end byte index 25405 is not a char boundary; it is inside '—' (bytes 25403..25406 of string)
```

The panic killed the agent actor task. From 06:20 UTC on 2026-08-30 to at least 12:20 UTC on 2026-08-31 (where the log excerpt ends, 30+ hours), **every** distill duty tick failed with:

```
Duty error (will retry next period): Agent shut down
duty: distill
```

— 24+ consecutive hourly failures. The daemon process stayed up the whole time (Telegram polling, channel loops, and other channels kept logging), but nothing was behind the mailbox: `AgentHandle::spawn` does `tokio::spawn(actor.run())` and drops the `JoinHandle` (src/agent/handle.rs:76). `send_message_with_role` maps the closed-oneshot error to `"Agent shut down"` (src/agent/handle.rs:134) and the duty layer dutifully retries next period, forever.

The *trigger* (the char-boundary panic in the file_edit success echo) was already fixed in 2a387d8, and 8d5442e denied the whole panic class (unwrap_used/expect_used/panic/string_slice) so this specific trigger cannot recur. But the failure mode is the finding: **one panicking tool invocation inside one duty turn permanently killed the actor for the whole daemon, and nothing noticed or recovered.** A future panic that escapes the lints (unsafe-adjacent code, allocation failure handling, a third-party crate panic, any `#[allow]`ed site) reproduces a 30-hour silent outage.

## Analysis

- `tokio::spawn(actor.run())` returns a `JoinHandle` that is discarded. A panic in `actor.run()` (or in any awaited tool future inside `handle()`) ends the actor loop; the mpsc receiver is dropped; every subsequent envelope send "succeeds" (channel not full) but the oneshot reply sender is dropped, so callers see the closed channel.
- The duty dispatcher's retry ("will retry next period") is correct behavior for transient failures but cannot distinguish "actor busy" from "actor dead". There is no liveness signal anywhere.
- A panic in a *duty-driven* turn killed the actor for *all* channels — Telegram, socket, GitHub, Linear share the one actor. The blast radius of a single tool panic is the whole bot.

## Suspected code location

- `src/agent/handle.rs:76` — `tokio::spawn(actor.run())`, `JoinHandle` dropped, no supervision.
- `src/duty/mod.rs` — duty dispatch treats "Agent shut down" as an ordinary retryable error; no escalation.
- `src/agent/handle.rs:134` — `"Agent shut down"` string is the only signal of actor death; it carries no recovery action.

## Possible directions (needs a design call)

1. **Supervise and restart**: keep the `JoinHandle`, monitor it, and on unexpected termination restart the actor with its persisted state (sessions/watermarks live in the workspace already). Simplest and matches how the duty layer already assumes "retry next period" is meaningful.
2. **Contain panics at the turn boundary**: wrap `self.handle(&input)` in `FutureExt::catch_unwind` inside the actor loop so a panicking turn degrades to an error reply instead of killing the mailbox owner.
3. Both: containment per turn, supervision as the backstop.

Option 2 alone changes panic semantics of the turn loop; option 1 alone still loses the in-flight turn on restart. This needs a decision on which guarantee matters before code.

amackillop:
Closes #156.

The daemon now observes the agent actor task while serving. An unexpected actor exit or panic terminates the process so systemd restarts it; SIGTERM drain behavior is unchanged.

Tests: `just check`.


kitaebot (approved):
Verdict: **correct** — the supervision does what the commit message says and the drain path is provably untouched.

The actor `JoinHandle` is polled by `&mut` inside the serving `tokio::select!`, so supervision of unexpected exit/panic costs the drain path nothing: on SIGTERM the other select arms (including the supervision arm) are dropped and the handle is still awaitable by value. Clean completion maps to `AgentExited`, join error to `AgentTaskFailed`, both exiting non-zero — and `Restart=on-failure` correctly ignores explicit stops. Drain-time panics now surface as a non-zero exit instead of a log line, which is the actual fix. The new test drives the panic path through the real `run_with_shutdown` plumbing; the clean-exit variant is untestable by construction (actor has no exit path) and the defensive variant is acceptable. Specs 01/09/24 and README match the behavior.

PR #167, opened 2026-09-18
Block piped test commands

amackillop:
## Summary

Reject piped `cargo test` and `just test*` invocations before filters can discard failure evidence needed by the #145 trailer.

## Validation

- `just test-one tools::exec`
- `just check`


kitaebot (approved):
Verdict: **correct** — the detector does what the commit message claims, with quoting semantics that survive contact with the shell.

The new check tracks quote state itself and correctly distinguishes an unquoted `|` from `||` control flow, resets candidates on `;`/`&`/newline, handles env-prefixed and absolute-path invocations, and stays literal on escaped pipes and quoted text — the #135-class false positive (`echo 'cargo test | grep failed'`) is covered by a test. It integrates first inside `command_blocked`, ahead of the per-segment deny rules, and the spec edit matches the implementation's actual behavior. Test coverage spans the block cases (env prefix, absolute path, the three `just test*` recipes) and the allow cases.

Residual evasions (`env`/`time` wrappers, pipes inside `$(…)`) are shared with the existing structural deny layer, not regressions of this PR.

